PR #48, opened 2024-03-13
Add SQL message database

Hyodar:
This PR:
* Does a minor refactoring on test utilities
* Removes `aggregator.MessageDatabase`
* Adds `core.database.Database` with a similar interface and plugs it into the Aggregator

On considerations about how the models are organized - keep in mind this is an SQLite database, so there are some restrictons such as not being able to store arrays in columns (in these cases simply `JSON` was used, which doesn't impact querying much as it's not used in clauses, and seems like the best choice because it can effectively be used in queries as opposed to a blob) or unsigned 64-bit integers (`text` by default to avoid loss, lossy `integer` for timestamps, which we can safely validate being in `int64` range when storing).

It would also be good to establish canonical message types so we can use one 'go-native' format for everything and just freely parse it into the model and contract binding types whenever needed, validating fields such as the timestamp. Though IMO this is work for another PR.


core/database/database.go
taco-paco, line 1: This database is related only to `aggregator`. There's no reason for it to be in core which `operator` has access to. If we will have some common interface in the future we can move it to `core` but implementations have to be in the corresponding folders

taco-paco (changes requested):
Looks good. The thing that I disagree with is moving database to `core`.

core/database/database.go
taco-paco: Since we discussed creating db for operator following this logic it has to be in core as well.

Hyodar: At the moment I don't see why operator and aggregator DBs would be different, that's why I moved it to `core`. Wdyt?

Hyodar: Anyway, I don't think it's really in the scope of this moving the db to core, so reverting that part.

taco-paco (approved):
Beautiful

PR #52, opened 2024-03-15
Fix integration tests on Mac

taco-paco:
Issues:
1. Prior messages that failed to be resend were dropped, hence never got to aggregator.
2. Port 5000 is occupied on Mac by Airplay

Introduced:
Rpc client now has a queue of unsent messages

1. All messages that failed to be send make it into queue.
2. On every new send we try to resend existing messages
3. The messages that failed to be send are remained in the queue
4. Try to send current message. In case failed see 1.

There's also a timer that occasionally triggers resend

### DIFF
--- a/operator/consumer/queues_listener.go
+++ b/operator/consumer/queues_listener.go
@@ -33,7 +33,7 @@ func NewQueuesListener(receivedBlocksC chan<- BlockData, logger logging.Logger) 
 	return listener
 }
 
-func (l *QueuesListener) Add(rollupId uint32, rollupDataC <-chan rmq.Delivery, ctx context.Context) error {
+func (l *QueuesListener) Add(ctx context.Context, rollupId uint32, rollupDataC <-chan rmq.Delivery) error {
 	l.queueDeliveryMutex.Lock()
 	defer l.queueDeliveryMutex.Unlock()
 
@@ -42,7 +42,7 @@ func (l *QueuesListener) Add(rollupId uint32, rollupDataC <-chan rmq.Delivery, c
 	}
 	l.queueDeliveryCs[rollupId] = rollupDataC
 
-	go l.listen(rollupId, rollupDataC, ctx)
+	go l.listen(ctx, rollupId, rollupDataC)
 
 	return nil
 }
@@ -53,7 +53,7 @@ func (l *QueuesListener) Remove(rollupId uint32) {
 	l.queueDeliveryMutex.Unlock()
 }
 
-func (l *QueuesListener) listen(rollupId uint32, rollupDataC <-chan rmq.Delivery, ctx context.Context) {
+func (l *QueuesListener) listen(ctx context.Context, rollupId uint32, rollupDataC <-chan rmq.Delivery) {
 	for {
 		select {
 		case d, ok := <-rollupDataC:
--- a/operator/consumer/queues_listener.go
+++ b/operator/consumer/queues_listener.go
@@ -3,6 +3,7 @@ package consumer
 import (
 	"context"
 	"errors"
+	"sync"
 
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/ethereum/go-ethereum/core/types"
@@ -15,8 +16,9 @@ var (
 )
 
 type QueuesListener struct {
-	receivedBlocksC chan<- BlockData
-	queueDeliveryCs map[uint32]<-chan rmq.Delivery
+	receivedBlocksC    chan<- BlockData
+	queueDeliveryCs    map[uint32]<-chan rmq.Delivery
+	queueDeliveryMutex sync.Mutex
 
 	logger logging.Logger
 }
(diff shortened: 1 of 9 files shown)


operator/rpc_client.go
```diff
@@ -42,82 +75,172 @@ func (c *AggregatorRpcClient) dialAggregatorRpcClient() error {
...
+
+		// Critically ugly section
+		{
+			c.unsentMessagesMutex.Lock()
+			if len(c.unsentMessages) == 0 {
+				c.unsentMessagesMutex.Unlock()
+				continue
+			}
+			c.unsentMessagesMutex.Unlock()
+		}
```
Hyodar, line 102:
```suggestion
		c.unsentMessagesMutex.Lock()
		unsentMessagesLen := len(c.unsentMessages)
		c.unsentMessagesMutex.Unlock()

		if unsentMessagesLen == 0 {
			continue
		}
```

operator/rpc_client.go
```diff
@@ -42,82 +75,172 @@ func (c *AggregatorRpcClient) dialAggregatorRpcClient() error {
...
-		time.Sleep(retryInterval)
+		if err != nil {
+			c.logger.Error("Couldn't resend message", "err", err)
+
+			c.unsentMessages[errorPos] = message
+			errorPos++
+		}
+	}
+
+	c.unsentMessages = c.unsentMessages[:errorPos]
```
Hyodar, line 156: I wonder if having a separate array for messages that couldn't be re-sent, which would be simpler, wouldn't even be better perf-wise. We're still resizing the original array here, which might trigger a realloc anyway, and range loops are usually more efficient. Not blocking, but I had to stop for some minutes here to make sure this was safe, so maybe it's a worthwhile change.

tests/integration/integration_test.go
```diff
@@ -840,3 +865,11 @@ func (ai *AnvilInstance) setBalance(address common.Address, balance *big.Int) er
 func (ai *AnvilInstance) mine(blockCount, timestampInterval *big.Int) error {
 	return ai.WsClient.Client.Client().Call(nil, "anvil_mine", "0x"+blockCount.Text(16), "0x"+timestampInterval.Text(16))
 }
+
+func (ai *AnvilInstance) getCurrentBlockHeight() (*big.Int, error) {
+	var result hexutil.Big
+	if err := ai.WsClient.Client.Client().Call(&result, "eth_blockNumber"); err != nil {
```
Hyodar, line 871: You could simply do `ai.WsClient.BlockNumber()` here - though this returns a `uint64`, so there are some more places to change this.

Btw, the only reason an RPC call was made manually above is because `anvil_mine` is not standard, it's specific from anvil.

Hyodar (approved):
GJ, please check the comments, but none are actually blocking

operator/consumer/queues_listener.go
```diff
@@ -15,8 +16,9 @@ var (
 )
 
 type QueuesListener struct {
-	receivedBlocksC chan<- BlockData
-	queueDeliveryCs map[uint32]<-chan rmq.Delivery
+	receivedBlocksC    chan<- BlockData
+	queueDeliveryCs    map[uint32]<-chan rmq.Delivery
+	queueDeliveryMutex sync.Mutex
```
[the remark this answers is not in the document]
taco-paco: Done

operator/consumer/queues_listener.go
```diff
@@ -31,39 +33,49 @@ func NewQueuesListener(receivedBlocksC chan<- BlockData, logger logging.Logger)
 	return listener
 }
 
-func (listener *QueuesListener) Add(rollupId uint32, rollupDataC <-chan rmq.Delivery, ctx context.Context) error {
-	if _, exists := listener.queueDeliveryCs[rollupId]; exists {
+func (l *QueuesListener) Add(rollupId uint32, rollupDataC <-chan rmq.Delivery, ctx context.Context) error {
```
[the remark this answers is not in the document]
taco-paco: Done

operator/rpc_client.go
```diff
@@ -16,24 +17,56 @@ type AggregatorRpcClienter interface {
...
+)
+
+type RpcMessageType struct {
+	MessageType int
+	Message     interface{}
+}
+
 type AggregatorRpcClient struct {
-	rpcClient            *rpc.Client
+	rpcClientMutex sync.Mutex
```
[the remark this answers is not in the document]
taco-paco: Done

tests/integration/integration_test.go
taco-paco: True, removed

operator/rpc_client.go
taco-paco: Tbh I don't mind 2nd array approach. Tho I would like to point out that the critical moment here is memory. In worst case there will be x2 of memory that array uses. With db we want to have a limit_size at which we going to dump them. So need to be careful here. 
Regarding reallocation I'm 99% sure it shalln't happen since we only decrease this of array - changing the header. So I will be very surprised if any reallocation happens. This is [standard algorithm ](https://en.cppreference.com/w/cpp/algorithm/remove)to avoid exactly that case.

operator/rpc_client.go
```diff
@@ -42,82 +75,172 @@ func (c *AggregatorRpcClient) dialAggregatorRpcClient() error {
...
+			err = c.rpcClient.Call("Aggregator.ProcessSignedCheckpointTaskResponse", signedCheckpointTaskResponse, &reply)
 
-		if err == nil {
-			c.logger.Info("Data successfully sent to aggregator")
-			return
+		case StateRootType:
+			signedStateRootUpdateMessage := message.Message.(*coretypes.SignedStateRootUpdateMessage)
+			err = c.rpcClient.Call("Aggregator.ProcessSignedStateRootUpdateMessage", signedStateRootUpdateMessage, &reply)
+
+		case OperatorSetUpdateType:
```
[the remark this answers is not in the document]
taco-paco: I didn't like the idea of passing raw interface into db methods so I done it this way. Don't have a strong preference tho

operator/rpc_client.go
taco-paco: Technically you risking of having actual `len` changed since its not in critical section anymore and someone else can jump in between

Hyodar: Those seem like equivalent approaches for me, though of course depends on what it turns into when compiling. How do you think they're differing?

Hyodar: I'm not sure we can assume it's critical session once the mutex is unlocked

operator/rpc_client.go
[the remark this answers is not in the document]
Hyodar: Sure, not a problem. What I mean is you're using a `MessageType` thing to do checks and then casting the raw interface. For example, instead of:
```go
		switch message.MessageType {
		case CheckpointType:
			signedCheckpointTaskResponse := message.Message.(*coretypes.SignedCheckpointTaskResponse)
			err = c.rpcClient.Call("Aggregator.ProcessSignedCheckpointTaskResponse", signedCheckpointTaskResponse, &reply)

		case StateRootType:
			signedStateRootUpdateMessage := message.Message.(*coretypes.SignedStateRootUpdateMessage)
			err = c.rpcClient.Call("Aggregator.ProcessSignedStateRootUpdateMessage", signedStateRootUpdateMessage, &reply)

		case OperatorSetUpdateType:
			signedOperatorSetUpdateMessage := message.Message.(*coretypes.SignedOperatorSetUpdateMessage)
			err = c.rpcClient.Call("Aggregator.ProcessSignedOperatorSetUpdateMessage", signedOperatorSetUpdateMessage, &reply)
}
```
You could do:
```go
switch msg := message.Message.(type) {
case *coretypes.SignedCheckpointTaskResponse:
    err = c.rpcClient.Call("Aggregator.ProcessSignedCheckpointTaskResponse", msg, &reply)
case *coretypes.SignedStateRootUpdateMessage:
    err = c.rpcClient.Call("Aggregator.ProcessSignedStateRootUpdateMessage", msg, &reply)
case *coretypes.SignedOperatorSetUpdateMessage:
    err = c.rpcClient.Call("Aggregator.ProcessSignedOperatorSetUpdateMessage", msg, &reply)
}
```

Just seems a lot better to me, and then we could drop the message type field and also the `RpcMessageType`.

operator/rpc_client.go
taco-paco: The thing is they in proposed approach you're not checking actual `len` of array, you checking `len` that was in critical section - snapshot at that moment. Since there's no lock anymore, another routine could take over, change `c.unsentMessages` and you would basing your decisions on incorrect `len`(`unsentMessagesLen `).

This is wrong way to check if array size is 0 in that particular moment

Hyodar: Yup, but both have the same problem. Again, the only operation that is relevant in terms of the critical session itself is getting the length. Concurrency can only affect the length, not the check. If immediately after the lock is released in the current implementation it was changed, the outcome would be exactly the same.

Hyodar: Btw, can't we simply start and stop the ticker to avoid useless iterations and avoid this double check altogether? This is already done on `tryResend` anyway.

Hyodar: Anyway, if we can't we can do any of the two approaches, np. In my opinion both work separately, but the truth is it's a workaround to avoid re-initializing the client.
Another option would be just doing it inside `tryResend` after the length check, seems better even.

Hyodar: If we are to keep the current version somehow, I do think the comment is unnecessary

taco-paco: > he only operation that is relevant in terms of the critical session itself is getting the length

The critical section here is making decision if we need to skip or not and for that you need a correct length.

This check exist for not triggering client initilization prematurely since it shall be done only on in first send.

> Concurrency can only affect the length, not the check

check depend on the length. In critical section its valid, outside not.
current way:
If we didn't `continue` in critical section then we don't care what current length is in context of initializing client. If it wasn't 0 we can try initializing.

proposed one.
The value in critical section was 0. Current one unknown. Basing skip on past

Hyodar: Keeping track of internal conversations - we'll keep the current approach for this PR and I'll just push another PR later removing this check and centralizing the RPC client initialization on `tryResendFromDeque`.

operator/rpc_client.go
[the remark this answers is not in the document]
taco-paco: Done

PR #57, opened 2024-04-03
Add checkpoint task attestation

Hyodar:


operator/rpc_client.go
```diff
@@ -216,3 +236,22 @@ func (c *AggregatorRpcClient) SendSignedOperatorSetUpdateToAggregator(signedOper
 		return nil
 	}, signedOperatorSetUpdateMessage)
 }
+
+func (c *AggregatorRpcClient) GetAggregatedCheckpointMessages(fromTimestamp, toTimestamp uint64, checkpointMessages *messages.CheckpointMessages) error {
```
taco-paco, line 240:
```suggestion
func (c *AggregatorRpcClient) GetAggregatedCheckpointMessages(fromTimestamp, toTimestamp uint64) (*messages.CheckpointMessages, error) {
```
Since its a getter lets just return `*messages.CheckpointMessages`.

There will be no need to do smth like this before calling then.
`var checkpointMessages messages.CheckpointMessages
`

this can be done inside the function itself

taco-paco:
```
func (c *AggregatorRpcClient) GetAggregatedCheckpointMessages(fromTimestamp, toTimestamp uint64) (*messages.CheckpointMessages, error) {
	c.logger.Info("Getting checkpoint messages from aggregator")

	type Args struct {
		FromTimestamp, ToTimestamp uint64
	}

	var checkpointMessages *messages.CheckpointMessages
	err := c.sendRequest(func() error {
		err := c.rpcClient.Call("Aggregator.GetAggregatedCheckpointMessages", &Args{fromTimestamp, toTimestamp}, checkpointMessages)
		if err != nil {
			c.logger.Info("Received error from aggregator", "err", err)
			return err
		}

		c.logger.Info("Checkpoint messages fetched from aggregator")
		return nil
	})

	return checkpointMessages, err
}
```

operator/operator.go
```diff
@@ -284,13 +284,35 @@ func (o *Operator) Start(ctx context.Context) error {
...
-		case checkpointTaskResponse, ok := <-checkpointTaskResponseChan:
+		case checkpointTaskCreatedEvent, ok := <-checkpointTaskCreatedChan:
 			if !ok {
 				o.logger.Info("Closing Operator")
 				return o.Close()
 			}
 
 			o.metrics.IncNumTasksReceived()
+
+			var checkpointMessages messages.CheckpointMessages
```
taco-paco, line 295: Let's move this to a separate function and call as goroutine since it has GetAggregatedCheckpointMessages request on which the whole loop would wait for response

taco-paco (approved):
Looks good. Just some minor comments

operator/rpc_client.go
taco-paco: Also noticed `databaser` fetch functions done in the same fashion. 
Such pattern usually applied when "filler" function agnostic to dst and just fills it, `void*` in C or` interface{}` in Go.

For example:
borsh: `func Deserialize(s interface{}, data []byte) error`
`void *memcpy(void *dest, const void * src, size_t n)`
In above the type needed specified outside the call but In cases of Fetch* functions or this one we already know a type. So there's no need for such semantic

operator/operator.go
Hyodar: Do you think it's that problematic? Considering checkpoints are done once every some hours and this should ideally not take too much time, I think we could just do it like this and avoid one extra channel. Wdyt?

Hyodar: Either way, the error management there is pretty simple, so making it a goroutine is simple as well

Hyodar: Done in [dae24b9](https://github.com/NethermindEth/near-sffl/pull/57/commits/dae24b966e94463a4f15273e5ae1074a0807385c)

operator/rpc_client.go
Hyodar: Done in [d32ac59](https://github.com/NethermindEth/near-sffl/pull/57/commits/d32ac59b490af6668132639963d4425c573560d8) and [36c7217](https://github.com/NethermindEth/near-sffl/pull/57/commits/36c721776cfde7b3012f6e463861e4ecba9c615c)

PR #63, opened 2024-04-09
Add pausing to relevant contracts

Hyodar:


taco-paco (commented):
Why is it necessary for contracts to be pausable?

Hyodar:
This is not a necessary change per se, just flexibility. If we have an issue in these operations somehow, ideally only in the context of testnet, we can pause them.

taco-paco (approved):

PR #64, opened 2024-04-10
Add operator whitelisting

Hyodar:


taco-paco (commented):
I mean code looks ok. But what is the purpose of this and how is this not a centralization?

Operator before joining needs to be included in whitelist

taco-paco:
This seems to be critical change for me I guess needs to be discussed with NEAR.

Hyodar:
This is the expected behavior for testnet for now. We'll have only a small number of operators we're closely working with. Has already been discussed with NEAR

taco-paco (approved):
LGTM. I would add a small TODO or note marking that its supposed to be removed

PR #65, opened 2024-04-10
Fix docker compose rollup contract deployment

Hyodar:


docker-compose.yml
```diff
@@ -86,12 +86,13 @@ services:
     command:
```
taco-paco, line 86: Let's move that magic into the file
```suggestion
    command:
      - -c
      - ./superscript.sh 
```

taco-paco (approved):

docker-compose.yml
Hyodar: Not sure if this is worth it now, with also the PR for initializing the operator set after deployment, I'm making a deployment script for this in another PR

PR #70, opened 2024-04-13
Improve and add deployment scripts

Hyodar:


Hyodar:
Merging this into #71. Will make another PR for the Holesky-specific scripts.

PR #101, opened 2024-04-20
Add more logs to MQ block processing and fix mutex copying

Hyodar:


operator/consumer/queues_listener.go
```diff
@@ -52,8 +52,8 @@ type QueuesListener struct {
 	logger logging.Logger
 }
 
-func NewQueuesListener(receivedBlocksC chan<- BlockData, logger logging.Logger) QueuesListener {
-	listener := QueuesListener{
+func NewQueuesListener(receivedBlocksC chan<- BlockData, logger logging.Logger) *QueuesListener {
```
taco-paco, line 55: Mutex copy occured here?

taco-paco (approved):

operator/consumer/queues_listener.go
Hyodar: Yup, and on `GetBlockStream`

taco-paco: Yeah, my bad. good catch

Hyodar: Well, if changing the queuesListener to a reference, as I did there, it would be okay. If the idea was keeping it as a value, since Consumer is a value there, it would be copied.
I made both changes because it seems bug-prone to in receivers sometimes treat the type as a value and sometimes as a reference. If we're adding something else that shouldn't be copied, we'd have another issue.

PR #102, opened 2024-04-21
Resend from queue on message send only if successful

Hyodar:
It seems excessive to try resending messages on every send, before the actual message is sent. If we know the original message send did not work, it doesn't make sense to retry for the enqueued ones.

### DIFF
--- a/operator/rpc_client.go
+++ b/operator/rpc_client.go
@@ -162,14 +162,14 @@ func (c *AggregatorRpcClient) sendOperatorMessage(sendCb func() error, message R
 		return
 	}
 
-	c.tryResendFromDeque()
-
 	c.logger.Info("Sending request to aggregator")
 	err = sendCb()
 	if err != nil {
 		appendProtected()
 		return
 	}
+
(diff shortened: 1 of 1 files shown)


taco-paco (approved):
This is ok if order isn't important. Did it cause any delays?

Hyodar:
The overall loop was taking longer when resending everytime, but it was probably not that significant. Before #100 it was worse, since all messages were being resent. As for the current state, we also have a tick, so I don't expect this to cause any issues in terms of delivery time.

taco-paco:
Ow, that was done in purpose since some of them failed due to TaskNotFoundErrorFn. And we don't have an ability to track those rn.  That was a problem especially with checkpoints since operator don't know anything about the fact if they were created or not. So this loops goes forward with resending

> "task %d not initialized or already completed"

taco-paco:
I think this is fine to merge, tho we need to add a proper error handle and be able to tell a part "already completed" and "not initialized"

Hyodar:
Makes sense, doing this later then.

PR #103, opened 2024-04-21
Start indexing from latest block synced

Hyodar:
Indexing from interruption was basically making it so it indexes *everything* from the latest interruption, effectively making it so the indexer would be behind the block rate for quite some time before actually reaching the current block.
This would be interesting if we made use of this capability somehow, but considering timeouts from MQ and message aggregation, if the indexer is ever stopped, the operator would already be inactive enough for this to be a fault and timeouts would be easily reached (startup takes quite a while).
In the end, keeping it like that only makes it so the operator doesn't realize there's a submission for current messages and has older messages to sign that he's already inactive for, which is what was happening in the testing environment.


taco-paco (approved):
🚀

PR #104, opened 2024-04-21
Add message prefixes

Hyodar:
This PR adds prefixes to messages and task responses with the objective of making it even more difficult to have collisions with other messaging formats. Not all AVSs do this, but it's a small cost for a huge benefit. Now, these hashes are `hash(prefix, hash(data))`, which also covers the possibility of a similar data format including a 32-byte field.
I'm following a simpler approach than EIP712 because in some cases we wouldn't benefit too much from the features, and in some cases, such as in `verifyingContract`, we plainly don't want them so we don't have a digest for every chain.


core/utils.go
```diff
@@ -19,6 +19,32 @@ func Keccak256(data []byte) ([32]byte, error) {
 	return digest, nil
 }
 
+func HashWithPrefix(prefix []byte, data []byte) ([32]byte, error) {
+	prefixHash, err := Keccak256(prefix)
```
taco-paco, line 23: Keccak shalln't return an error.

taco-paco: Implementation also doesn't

taco-paco (approved):

core/utils.go
Hyodar: Indeed - checking this out

Hyodar: Done in [e9c4928](https://github.com/NethermindEth/near-sffl/pull/104/commits/e9c4928ff7f1c5b675af7c592a98376a0f50d085)

PR #110, opened 2024-04-22
Update eigen-sdk to 0.1.4

Hyodar:
Previous version is the one EigenDA uses in their Holesky branch, but this one is still based on Holesky. Still need to double-check if this does not break anything and update the message BLS aggregation service to use the new types, but seems to work.
At the current stage, though, putting this on the side for a while to avoid unexpected impacts.


Hyodar:
Yes, it is. Though as mentioned there I'll probably leave this one on hold for a bit, just so we're more stable for now.

PR #112, opened 2024-04-22
Increase block data buffer on notifier

Hyodar:
We were basically reaching a 'deadlock'-like state because the queue was full and as such the MQ listener was not being able to listen anymore.

### DIFF
--- a/operator/attestor/notifier.go
+++ b/operator/attestor/notifier.go
@@ -28,7 +28,7 @@ func (notifier *Notifier) Subscribe(rollupId uint32) (<-chan consumer.BlockData,
 		notifier.rollupIdsToSubscribers[rollupId] = list.New()
 	}
 
-	notifierC := make(chan consumer.BlockData, 10)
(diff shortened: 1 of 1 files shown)


taco-paco (approved):

taco-paco (approved):
I guess it started happening after batching blocks on relayer. I knew that it could deadlock but didn't expect it would. 

We need to control it on relayer side as well limiting batch with 100 blocks. Otherwise with unliky context switches we can hit that.

When I was writing that I had idea with callbacks or filling candidate queue that would be checked with some timer on attestor side. 

I guess good solution would be passing predicate to Notifier which would be checked in that loop and then notifying only correct target. I guess this is the way to go

Hyodar:
Agreed. Atm, that is not really feasible to reach considering the timeouts on transactions and block rates, but we need to add some logic there ideally to limit it - should be a bit lower than the channel buffer limit since this also depends on the consumption rate.

PR #115, opened 2024-04-22
Also resubmit blocks to NEAR DA on expired transaction

Hyodar:
After quite some execution time, there was this error: `NEAR DA client handler error: [An error happened during transaction execution: Expired]`. Transactions expiring are also a possibility, so including them in the resubmission logic.

### DIFF
--- a/relayer/relayer.go
+++ b/relayer/relayer.go
@@ -2,6 +2,7 @@ package relayer
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"time"
 
@@ -18,6 +19,7 @@ const (
 	NAMESPACE_ID               = 1
 	SUBMIT_BLOCK_INTERVAL      = 1500 * time.Millisecond
 	SUBMIT_BLOCK_RETRY_TIMEOUT = 1 * time.Second
+	SUBMIT_BLOCK_RETRIES       = 3
 )
 
 type Relayer struct {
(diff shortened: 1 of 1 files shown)


relayer/relayer.go
```diff
@@ -88,7 +88,10 @@ func (r *Relayer) Start(ctx context.Context) error {
 					}
 				}
 
-				return err
+				if err != nil {
```
taco-paco, line 91: Ideally that should also check case above. Then one more time if same error and so on
`if strings.Contains(err.Error(), "InvalidNonce") || strings.Contains(err.Error(), "Expired"`

taco-paco (approved):

relayer/relayer.go
Hyodar: We can add some retries - I don't think I've ever seen it reach an invalid nonce and not recover in the next step though. Point is it's not necessary it'll be too many otherwise the MQ timeout will be reached anyway.

Hyodar: Done in [2672c4c](https://github.com/NethermindEth/near-sffl/pull/115/commits/2672c4c2a6c171376d4f8f0a8e8df87e2350cb29)

taco-paco: >  Point is it's not necessary it'll be too many otherwise the MQ timeout will be reached anyway.
It will be reached anyway if relayer fails to send. Relayer can't also handle discarding of the message since it doesn't know state of operator. It's business just to send.

PR #129, opened 2024-04-28
Increase relayer intervals

Hyodar:
We're probably overestimating NEAR's throughput here. Even with these changes, transactions are still expiring (after 3 retries) frequently. Still going to try increasing the usual interval a bit more and also retries to check this out, as it's unclear whether this is affected by our transaction rate or not. Maybe we need to push this through another RPC besides the default one?

Even on a retry it's still nowhere near 100, but I also increased notifier's channel queue size for some additional safety - we can probably increase it quite a bit more still.

### DIFF
--- a/operator/attestor/notifier.go
+++ b/operator/attestor/notifier.go
@@ -28,7 +28,7 @@ func (notifier *Notifier) Subscribe(rollupId uint32) (<-chan consumer.BlockData,
 		notifier.rollupIdsToSubscribers[rollupId] = list.New()
 	}
 
-	notifierC := make(chan consumer.BlockData, 100)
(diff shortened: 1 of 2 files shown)


operator/attestor/notifier.go
```diff
@@ -28,7 +28,7 @@ func (notifier *Notifier) Subscribe(rollupId uint32) (<-chan consumer.BlockData,
 		notifier.rollupIdsToSubscribers[rollupId] = list.New()
 	}
 
-	notifierC := make(chan consumer.BlockData, 100)
```
taco-paco, line 31: Shall be resolved with this one
https://github.com/NethermindEth/near-sffl/pull/131/files

taco-paco (approved):

PR #131, opened 2024-04-29
feat: add predicate to filter destination

taco-paco:


operator/attestor/notifier.go
```diff
@@ -7,6 +7,12 @@ import (
 	"github.com/NethermindEth/near-sffl/operator/consumer"
 )
 
+type Predicate = func(consumer.BlockData) bool
```
Hyodar, line 10: `Predicate` seems a bit too general, something like `BlockPredicate` would be better IMO.

Hyodar: Also, why is this public and `subscriberData` isn't?

operator/attestor/notifier.go
```diff
@@ -20,18 +26,21 @@ func NewNotifier() Notifier {
...
+		predicate: predicate,
+		notifierC: make(chan consumer.BlockData, 100),
+	}
+	id := notifier.rollupIdsToSubscribers[rollupId].PushBack(subscriber)
 
-	return notifierC, id
+	return subscriber.notifierC, id
 }
 
 func (notifier *Notifier) Notify(rollupId uint32, block consumer.BlockData) error {
```
Hyodar, line 46: Shouldn't we be using a read lock only here? The predicate shouldn't take long though. Maybe we could also call `Notify` as a goroutine if that's cheap enough.

Hyodar (approved):
LGTM, just a few comments.

operator/attestor/notifier.go
taco-paco: `Predicate` is part of public API and `subscriberData` is only used within the file

operator/attestor/notifier.go
taco-paco: The RWMutex was discussed [prior](https://github.com/NethermindEth/near-sffl/pull/18#discussion_r1483875507). 

Regarding goroutine in Notify I guess we need to benchmark here.
1. If Notify execution waits for mutex for too long and if its a problem for `processMQBlocks`
2. If spawning goroutines for that is less expensive 

I feel like Notifies come in batches while Subscribe & Unsubscribe are more evenly distributed. So it can result us spawning bunch of goroutines in one shot not sure it will be more performant. 

Anyway... to be honest don't know what is better here, need empirical data to decide

operator/attestor/notifier.go
Hyodar: Alright, that's okay. It can be used outside

operator/attestor/notifier.go
Hyodar: Considering this is executed synchronously prob not much of a difference anyway, but otherwise, comparing the number of times notify is called vs subscribe I do think it would make sense to only read lock here

Hyodar: Yeah, no need to do this async if we don't need to. Seems to be working fine atm, so let's only keep this in mind

PR #135, opened 2024-04-29
Check shutdown error on resend

Hyodar:


operator/rpc_client.go
```diff
@@ -150,8 +150,8 @@ func (c *AggregatorRpcClient) tryResendFromDeque() {
 		if err != nil {
 			c.logger.Error("Couldn't resend message", "err", err)
 
-			if i == 0 {
-				c.logger.Error("Couldn't resend first message, most likely a connection error")
+			if i == 0 && err == rpc.ErrShutdown {
```
taco-paco, line 153: Feels like of we have `rpc.ErrShutdow` it doesn't really matter if `i==0`

taco-paco (approved):

operator/rpc_client.go
Hyodar: Problem is with how the queue management is working there, it does

taco-paco: ? not sure what you mean. Maybe you meant

`if i == 0 || err == rpc.ErrShutdown {`

Hyodar: If we're in between the queue, some messages were sent and some messages did throw anyway and at some point it shuts down, then the queue is 'dirty', we can't just leave early. We'd need to at least clean the queue.

taco-paco: We leave in queue messages that throw smth. If we are in the middle of the queue and we got shutdown error. We simply run smth like this

```
	for j := i; j < len(c.unsentMessages); j++ {
		message := c.unsentMessages[I]
		c.unsentMessages[errorPos] = message
               errorPos++
      }
       // return
```
    Evectively we say all messages after shutdown are unsent and shift them. Or there's smth else that is dirty?

Hyodar: Exactly, this is what I mean by cleaning. Didn't bother doing it because we'd be rewriting this once properly handling errors really soon (thought it'd be done already actually)

taco-paco: Wouldn't this be part of code anyway? I mean shutdown can happen even with proper errors from server side

Hyodar: Done in [ce3f507](https://github.com/NethermindEth/near-sffl/pull/135/commits/ce3f507b28521ef5473b3cb4d0060805ac5b3ef3)

taco-paco (approved):

PR #136, opened 2024-04-29
Only respond to task in the next block

Hyodar:
Because of how the EL BLS aggregation contracts work, you can't respond to a task in the same block it was created.

### DIFF
--- a/core/chainio/avs_writer.go
+++ b/core/chainio/avs_writer.go
@@ -2,6 +2,7 @@ package chainio
 
 import (
 	"context"
+	"time"
 
 	"github.com/NethermindEth/near-sffl/core/config"
 	"github.com/NethermindEth/near-sffl/core/types/messages"
@@ -107,6 +108,18 @@ func (w *AvsWriter) SendAggregatedResponse(
 	taskResponse messages.CheckpointTaskResponse,
 	aggregation messages.MessageBlsAggregation,
 ) (*types.Receipt, error) {
+	// Wait a block if the task TaskCreatedBlock is the same as the current block
+	currentBlock, err := w.client.BlockNumber(ctx)
+	if err != nil {
+		w.logger.Errorf("Error getting current block number")
+		return nil, err
+	}
+
+	if uint64(task.TaskCreatedBlock) == currentBlock {
+		w.logger.Info("Waiting roughly a block before sending aggregated response...")
+		time.Sleep(20 * time.Second)
+	}
+
 	txOpts, err := w.TxMgr.GetNoSendTxOpts()
 	if err != nil {
 		w.logger.Errorf("Error getting tx opts")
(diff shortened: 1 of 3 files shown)


core/chainio/avs_writer.go
```diff
@@ -107,6 +109,18 @@ func (w *AvsWriter) SendAggregatedResponse(
 	taskResponse messages.CheckpointTaskResponse,
 	aggregation messages.MessageBlsAggregation,
 ) (*types.Receipt, error) {
+	// Wait a block if the task TaskCreatedBlock is the same as the current block
+	currentBlock, err := w.client.BlockNumber(ctx)
```
taco-paco, line 113: The SendAggregatedResponse shalln't be triggered at the first place if conditions for writing aren't met.

I propose extracting this `sendAggregatedResponseToContract`

core/chainio/avs_writer.go
```diff
@@ -65,15 +66,16 @@ func BuildAvsWriter(txMgr txmgr.TxManager, registryCoordinatorAddr, operatorStat
...
 }
 
-func NewAvsWriter(avsRegistryWriter avsregistry.AvsRegistryWriter, avsServiceBindings *AvsManagersBindings, logger logging.Logger, txMgr txmgr.TxManager) *AvsWriter {
+func NewAvsWriter(avsRegistryWriter avsregistry.AvsRegistryWriter, avsServiceBindings *AvsManagersBindings, client eth.Client, logger logging.Logger, txMgr txmgr.TxManager) *AvsWriter {
 	return &AvsWriter{
 		AvsRegistryWriter:   avsRegistryWriter,
 		AvsContractBindings: avsServiceBindings,
-		logger:              logger,
 		TxMgr:               txMgr,
+		client:              client,
```
taco-paco, line 77: Feels like an excessive member just for on function

core/chainio/avs_writer.go
Hyodar: Could be, np with that

core/chainio/avs_writer.go
Hyodar: It was already there, just never initialized

core/chainio/avs_writer.go
Hyodar: Done in [bd3ab42](https://github.com/NethermindEth/near-sffl/pull/136/commits/bd3ab42a54b665762050a99dca7fe3bc992656eb)

core/chainio/avs_writer.go
Hyodar: Removed it in [4629ef4](https://github.com/NethermindEth/near-sffl/pull/136/commits/4629ef418d7ba2ac1f3852a0cc31bad62d0898ee)

taco-paco: > It was already there, just never initialized

It was passed as argument and propagated to struct that actually needs it to make txs. We just use those structs

taco-paco (approved):

core/chainio/avs_writer.go
Hyodar: Yeah, what I mean is it was a field of `AvsWriter`

PR #137, opened 2024-05-01
Integrated metrics into relayer

taco-paco:


relayer/relayer.go
```diff
@@ -3,6 +3,8 @@ package relayer
 import (
 	"context"
 	"errors"
+	"github.com/NethermindEth/near-sffl/core"
```
Hyodar, line 6: These imports could be separated into std, external and internal

relayer/relayer.go
```diff
@@ -43,14 +48,25 @@ func NewRelayerFromConfig(config *config.RelayerConfig, logger sdklogging.Logger
...
 		rpcClient:   rpcClient,
 		rpcUrl:      config.RpcUrl,
 		daAccountId: config.DaAccountId,
 		nearClient:  nearClient,
+		logger:      logger,
+		listener:    &SelectiveListener{},
 	}, nil
 }
 
+func (r *Relayer) WithMetrics(registry *prometheus.Registry) error {
```
Hyodar, line 60: Most times I've seen this `With` pattern in go it's used as a Builder, so it returns the object itself. Since this is fallible and not the same use-case, we may want to rename this.

relayer/metrics.go
```diff
@@ -0,0 +1,173 @@
...
+
+	if err := registry.Register(submissionDuration); err != nil {
+		return nil, fmt.Errorf("error registering submissionDuration histogram: %w", err)
+	}
+
+	return &SelectiveListener{
+		OnBlockReceivedCb: func() {
+			numBlocksReceived.Inc()
+		},
+		OnDaSubmissionFailedCb: func() {
```
Hyodar, line 101: Could you metrics for retries, expirations and invalid nonces?

relayer/cmd/main.go
```diff
@@ -15,52 +18,92 @@ import (
...
+	config := config.RelayerConfig{
+		Production:        ctx.Bool("production"),
+		RpcUrl:            ctx.String("rpc-url"),
+		DaAccountId:       ctx.String("da-account-id"),
+		KeyPath:           ctx.String("key-path"),
+		Network:           ctx.String("network"),
+		MetricsIpPortAddr: ctx.String("metrics-ip-port-address"),
 	}
 
+	fmt.Println("config.RpcUrl:", config.RpcUrl)
```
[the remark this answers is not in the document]
Hyodar: Also, non-submitted blocks would be good as well.

setup/aggregator/docker-compose.yml
```diff
@@ -1,11 +1,17 @@
 version: "3"
 
+networks:
+  near-sffl-aggregator:
```
Hyodar, line 4: I'd prefer if this had a different name to avoid confusion. Maybe a `-net` suffix?

setup/relayer/docker-compose.yml
```diff
@@ -1,12 +1,21 @@
 version: "3"
 
+networks:
+  near-sffl-relayer:
+    driver: bridge
+
 services:
   relayer_421614:
-    image: ghcr.io/nethermindeth/near-sffl/relayer:${SFFL_RELEASE}
+    image: near-sffl-test-relayer:latest
```
Hyodar, line 9: We should keep the release var

setup/relayer/docker-compose.yml
```diff
@@ -24,11 +35,16 @@ services:
         compress: "true"
 
   relayer_11155420:
-    image: ghcr.io/nethermindeth/near-sffl/relayer:${SFFL_RELEASE}
+    image: near-sffl-test-relayer:latest
```
Hyodar, line 38: Should also keep the var here

config-files/relayer1-docker-compose.yaml
Hyodar, line 1: This is not a docker compose file

Hyodar: Pls note the renaming wherever it's used as well

Hyodar (approved):
Looks good, nothing major, just a few comments

relayer/cmd/main.go
```diff
@@ -15,52 +18,92 @@ import (
...
-		cli.StringFlag{
-			Name:  "network",
-			Value: "http://127.0.0.1:3030",
-			Usage: "Network for NEAR client to use (options: Mainnet, Testnet, Custom url, default: http://127.0.0.1:3030)",
-		},
-	}
 	app.Name = "sffl-test-relayer"
 	app.Usage = "SFFL Test Relayer"
 	app.Description = "Super Fast Finality testing service that reads block data from an EVM network and feeds it to a NEAR DA contract."
+	app.Commands = []cli.Command{
```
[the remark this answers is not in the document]
taco-paco: Don't see anything like that. Would have to do that manually.

config-files/relayer1-docker-compose.yaml
taco-paco: Name was chosen analogous to other configs. They aren't docker compose files

relayer/relayer.go
taco-paco: Done

config-files/relayer1-docker-compose.yaml
taco-paco: I guess they're named so to show that they are used in docker compose

setup/aggregator/docker-compose.yml
taco-paco: Confusion with what?

relayer/metrics.go
taco-paco: What is expirations and how is it different from existing `numDaSubmissionsFailed`. Added other 2

relayer/cmd/main.go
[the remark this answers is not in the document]
taco-paco: Changed to run-*

relayer/cmd/main.go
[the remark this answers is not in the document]
taco-paco: Removed

> Also, non-submitted blocks would be good as well.

not sure what you mean

setup/relayer/docker-compose.yml
taco-paco: Reverted

relayer/cmd/main.go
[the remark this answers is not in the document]
Hyodar: Blocks that were not submitted after retries

relayer/metrics.go
Hyodar: When a transaction expires, from our error handling

setup/aggregator/docker-compose.yml
Hyodar: The container name

config-files/relayer1-docker-compose.yaml
Hyodar: Oh, that's okay, thought it was `relayer-1.docker-compose.yaml`

setup/aggregator/docker-compose.yml
Hyodar: Same applies to relayer btw

taco-paco: Don't really see in what scenario there shall be a confusion.

```
networks:
  near-sffl-aggregator:
    driver: bridge
```
In here it's obvious that ist network. Working from cli requires --network flag or network subcommand. 
It sounds smth like "wet water"

taco-paco: I prefer leaving it the way it is. Can change if it will be confusing indeed in some scenario

Hyodar: Well, it gets weird imo since there's already a container with the exact same name, and the network is not used only on it

Hyodar: But feel free to leave as it is, it's not blocking

relayer/metrics.go
Hyodar: There can be multiple reasons for failing to submit - currently we are categorizing them as invalid nonce, expired or unknown. This being the case, we can't really get the expirations with these two metrics

taco-paco: I see now, Done

PR #141, opened 2024-05-02
Change container name address to IP

Hyodar:


taco-paco (approved):
Could you verify that prometheus scrapes metrics successfully with this?

Hyodar:
Yes, I could

PR #146, opened 2024-05-06
Downgrade anvil foundry container version

Hyodar:
It seems like after the latest release of the foundry container, the block encoding on the integration test broke. This is really strange, going a bit deeper into it. In the meanwhile, downgrading it.


Hyodar:
Merging this earlier so we have int tests.

PR #151, opened 2024-05-07
Reinitialize Attestor client indefinitely

Hyodar:


operator/attestor/attestor.go
```diff
@@ -200,22 +201,36 @@ func (attestor *Attestor) reconnectClient(rollupId uint32) (eth.Client, error) {
...
+	reinitializeTicker := time.NewTicker(REINITIALIZE_DELAY)
+	reinitializeTicker.Stop()
+
+	reinitializeSubscription := func() error {
+		client, err := attestor.reconnectClient(rollupId)
+		if err != nil {
+			attestor.logger.Error("Error while reconnecting client", "rollupId", rollupId, "err", err)
+			return err
+		}
+		attestor.clients[rollupId] = client
```
taco-paco, line 213: Looking at it now it seems like we could use mutex here. If error happens on our side I guess multiple subscriptions could get it at the same time

taco-paco (approved):

operator/attestor/attestor.go
Hyodar: Yeah, makes sense. Those should be different clients, but it could happen multiple are triggered at once

Hyodar: Done in 8d7058c

PR #152, opened 2024-05-07
Add incremental message aggregation

Hyodar:


aggregator/message_blsagg.go
```diff
@@ -164,26 +174,71 @@ func (mbas *MessageBlsAggregatorService) singleMessageAggregatorGoroutineFunc(
...
+
+			aggregation := mbas.getMessageBlsAggregationResponse(messageDigest, validationInfo, false)
+
+			if aggregation.Status == types.MessageBlsAggregationStatusThresholdNotReached {
+				continue
+			}
+
+			mbas.aggregatedResponsesC <- aggregation
+
+			if aggregation.Status == types.MessageBlsAggregationStatusThresholdReached && !thresholdReached {
```
taco-paco, line 218: I feel like this can be simplified by splitting in 2 functions.

So here we have two stages:
1. Threshold not met yet. We collect within time span of `messageExpiredTimer`.
2. Threshold met before `messageExpiredTimer`. So we let people to still send until `thresholdReachedTimer`

So let's just call another function remnantAggragation in here and then exit.

taco-paco: IMO this looks much simpler. You get rid of overcomplicated timer management & thresholdReached that glues all together just to make to work.

```
func (mbas *MessageBlsAggregatorService) remnantAggregation(
	messageDigest coretypes.MessageDigest,
	quorumNumbers []eigentypes.QuorumNum,
	quorumThresholdPercentages []eigentypes.QuorumThresholdPercentage,
	signedMessageDigestsC <-chan SignedMessageDigest,
	aggregationTimeout time.Duration,
	ethBlockNumber uint64,
) {
	validationInfo := mbas.fetchValidationInfo(quorumNumbers, quorumThresholdPercentages, ethBlockNumber)
	thresholdReachedTimer := time.NewTimer(aggregationTimeout)

	for {
		select {
		case signedMessageDigest := <-signedMessageDigestsC:
			mbas.logger.Debug("Message goroutine received new signed message digest", "messageDigest", messageDigest)

			err := mbas.handleSignedMessageDigest(signedMessageDigest, validationInfo)
			signedMessageDigest.SignatureVerificationErrorC <- err
			if err != nil {
				continue
			}

			aggregation := mbas.getMessageBlsAggregationResponse(messageDigest, validationInfo, false)
			if aggregation.Status == types.MessageBlsAggregationStatusThresholdNotReached {
				continue
			}

			mbas.aggregatedResponsesC <- aggregation
			if aggregation.Status == types.MessageBlsAggregationStatusFullStakeThresholdMet {
				return
			}
			
		case <-thresholdReachedTimer.C:
			mbas.aggregatedResponsesC <- mbas.getMessageBlsAggregationResponse(messageDigest, validationInfo, true)
			return
		}
	}
}
```

taco-paco (commented):
Looks good, proposed some improvements for more transparent workflow

aggregator/message_blsagg.go
Hyodar: Sure, a second phase like that would work. Not exactly like that though, we'd need to share `validationInfo` instead of rebuilding it, but it makes timer management a bit better.

Hyodar: Done in [770b157](https://github.com/NethermindEth/near-sffl/pull/152/commits/770b157db37a646c1ce05fb346803a8b941ab177), [0e48296](https://github.com/NethermindEth/near-sffl/pull/152/commits/0e482962bfc82cc111ad257155ca5705ebd0d6ec) and [fa93080](https://github.com/NethermindEth/near-sffl/pull/152/commits/fa930802cd6b2e278bb598a03bd8700e12fd0290)

taco-paco (approved):
Looks great now!

PR #157, opened 2024-05-13
feat: non-blocking block_listener + expirable messages

taco-paco:


indexer/src/main.rs
```diff
@@ -59,6 +60,7 @@ fn run(home_dir: std::path::PathBuf, config: RunConfigArgs) -> Result<()> {
         }
         rmq_publisher.run(validated_stream);
 
+        // TODO: block_handle wether cancelled or Panics. Can handle
```
Hyodar, line 63: Didn't get that

Hyodar (approved):
LGTM

indexer/src/block_listener.rs
```diff
@@ -75,42 +81,127 @@ impl BlockListener {
         }
     }
 
+    async fn ticker(
```
therustmonk, line 84: Two essential recommendations:

1. Try to use composition, and avoid using associated functions. Add like the `TickerProcessor` struct and get access to all its data.
2. Try completely avoid `continue`, and `return` keywords. They prevent you from getting benefits of expression-based approach, that will help you refactor everything easily. It will help you think more about Rust as Ocaml/F#/Haskell than as C/Go/JS.
3. ⭐ I also avoid using fully/partly qualified names (except the `log` crate, and channels mods). So `Instant::now()` is better than `time::Instant::now()`. This allows you to keep the code more accurate.

For example, you can replace `ticker` and `flush` with the following struct composition (it compiles, but I haven't tried to keep 100% correctness, review is required).

```rust
struct TickerProcessor {
    done: Option<oneshot::Receiver<()>>,
    queue_protected: types::ProtectedQueue<ExpirableCandidateData>,
    candidates_sender: mpsc::Sender<CandidateData>,
    listener: Option<BlockEventListener>
}

impl TickerProcessor {
    async fn entrypoint(mut self) -> Result<(), Error> {
        let done = self.done.take().ok_or_else(|| Error::msg("no receiver"))?;
        // mapping futures + future::select + into_inner() call could be used instead
        select! {
            res = self.tick_loop() => { res }
            _ = done => { Ok(()) }
        }
    }

    async fn tick_loop(&mut self) -> Result<(), Error> {
        #[cfg(not(test))]
        const FLUSH_INTERVAL: Duration = Duration::from_secs(1);
        #[cfg(test)]
        const FLUSH_INTERVAL: Duration = Duration::from_millis(100);
        let mut interval = time::interval(FLUSH_INTERVAL);
        loop {
            interval.tick().await;
            self.flush().await?;
            let mut queue = self.queue_protected.lock().await;
            self.listener.as_ref().map(|l| l.current_queued_candidates.set(queue.len() as f64));
            interval.reset();
        }
    }

    async fn flush(&mut self) -> Result<bool, Error> {
        let mut queue = self.queue_protected.lock().await;
        if !queue.is_empty() {
            info!(target: INDEXER, "Flushing");
            let now = time::Instant::now();
            while let Some(candidate) = queue.front() {
                if now.duration_since(candidate.timestamp) >= EXPIRATION_TIMEOUT {
                    queue.pop_front();
                } else {
                    self.candidates_sender.try_send(candidate.clone().into())?;
                }
            }
            Ok(true)
        } else {
            Ok(false)
        }
    }
}
```

taco-paco: Thanks for the comment! There's actually another branch with very similiar idea. https://github.com/NethermindEth/near-sffl/tree/feat/indexer/manager.

Want to make a separate PR for this. Glad that there's a consensus on that implementation. Will tag you in PR once it's out.

PR #162, opened 2024-05-15
RMQ reconnections

taco-paco:


Hyodar (approved):

emlautarom1 (approved):
LGTM.

It would be nice if there were some tests for the `run` loop though.

indexer/src/rmq_publisher.rs
```diff
@@ -95,11 +99,72 @@ impl RabbitPublisher {
...
+        receiver: mpsc::Receiver<PublishData>,
+    ) -> Self {
+        Self {
+            connection_pool,
+            receiver,
+            listener,
+        }
+    }
+
+    pub async fn run(mut self) {
```
therustmonk, line 143: Yeah, this approach is preferred, to have a struct, an entrypoint method, and extra methods to handle particular cases.

therustmonk (approved):

Issue #163, opened 2024-05-16
Docker compose: container is unhealthy


emlautarom1:
After building all Docker images on a clean repo, running `docker compose up` results in an error after ~15 seconds: `dependency failed to start: container mainnet-anvil is unhealthy`

The issue seems to be due to the following healtcheck

https://github.com/NethermindEth/near-sffl/blob/61395b7fb789cf6c5e3ecce2af572926c8d6b19e/docker-compose.yml#L37

If you attach to the `mainnet-anvil` container logs you can see that the application is working as intended, and if you run the same health check command from your host machine (`nc -zv localhost 8545`) you can see that the service is reachable (this is due to the mapping `8545:8545` from the container to the host).

Running a shell inside the container shows the root of the problem: attach to the container using `docker exec -it mainnet-anvil /bin/sh`, run `nc -zv localhost 8545` and the exit code (`echo $?`) should be a `1` instead of a `0`.

The issue then seems to be with the command used to perform the healthcheck.

### PC Specs
- OS: Fedora 39 Workstation x86_64
- CPU: Ryzen 5 3600
- Docker version: 24.0.5

Hyodar:
Curiously enough this isn't a problem here. I can follow the same steps, but `nc` does have a `0` exit code. Still, `0.0.0.0` works as well, which makes sense. So, no problem with the approach in #164.

PR #164, opened 2024-05-16
Prefer `0.0.0.0` over `localhost`

emlautarom1:
Solves #163 by using `0.0.0.0` over `localhost` in healthchecks.


Hyodar (approved):
As mentioned in #163, it's okay on my side as well. LGTM!

PR #165, opened 2024-05-17
refactor: set rollup-da to near original repo. go mod tidy

taco-paco:


Hyodar:
Noticed the blobs are not being properly decoded on int test. This is most likely the reason why: https://github.com/near/rollup-data-availability/commit/f771f366d73a1be985a87d1310d72d96cb9aee8c. Now the blobs are only data and the blobs we were using prior to that are `LegacyBlob`s. This is not a trivial change because we then need to manage getting the commitment from somewhere (or maybe use just the transactionId?) and coordinate an update.
Please check the possible approaches here and let's keep this open a bit more.

Hyodar (changes requested):
As mentioned above.

taco-paco:
The issue is fixed having commitment computed on the indexer side. I assume near-da contract will be updated in the future and maybe will include some commitment. Will see.

indexer/src/types.rs
```diff
@@ -23,3 +27,29 @@ impl fmt::Display for CandidateData {
...
+        Self {
+            commitment,
+            data: value.data
+        }
+    }
+}
+
+pub(crate) fn try_transform_payload(payload: &[u8]) -> Result<Vec<u8>> {
+    let submit_request: SubmitRequest = borsh::de::from_slice(payload)?;
+    let commited_blob: CommitedBlob = Blob::new(submit_request.data).into();
```
Hyodar, line 53: commited -> committed

Hyodar:
What I don't like too much about this is that it's breaking, and integrating this would require full coordination between nodes and relayer. Even though we are under development, it'd be a lot better if we just try decoding it as a legacy blob and, if it doesn't work, we decode it as the current blob format, also keeping compatibility in the MQ data.

taco-paco:
> Overall everything seems great, what I don't like too much about this is that it's breaking
I don't think we shall be concerned at this point about backwards compatibility. It will overcomplicate the logic which is IMO unnecessary right now.

> integrating this would require full coordination between nodes and relayer
It will require at least updating relayer, otherwise there's no sense in this PR.

> decoding it as a legacy blob and, if it doesn't work, we decode it as the current blob format, also keeping compatibility in the MQ data.
If you imply relayer-indexer side. It would mean that relayer didn't update and then there's no point in this pr, If this is about indexer-operator I would prefer avoid these complications since we can at this stage

Hyodar:
Sadly with a running environment we need to worry about compatibility. It's not easy to coordinate updates, so breaking updates should be either avoided and scheduled or done in a way that's smooth - e.g. once every operator is running the backwards-compatible version, we can introduce the new version and nothing breaks. For this I do think we can do it with not much of an issue, so I'd prefer the latter approach.

taco-paco:
It’s not only about this PR but also about other breaking changes like this. If you choose migrations for when they can be avoided you will have a spaghetti code from one version to another. Right now there’s no reason for that. This pr can be postponed until scheduled update and then merged

Hyodar:
I agree, I don't think this one is urgent. Though my understanding is there can be a specific judgment on whether something needs to be breaking or not or how big the impact of a compatibility layer is. If it does result in a big rework, then not worth it, otherwise being a small adjustment it's better to advance with the feature and drop backward compatibility later.

PR #168, opened 2024-05-20
Format rollupId as decimal in aggregator metrics

Hyodar:
Quick fix - rollup IDs being formatted as hex in aggregator metric labels.


Hyodar:
Oops, agreed. The 'fix' was because for the operator metrics they were being formatted as decimals, and for aggregators hex. Let me change the title.

Issue #170, opened 2024-05-20
Add support for Operator `opt-out`


emlautarom1:
Currently, there is no way for an operator to "opt out" once it has "opted-in":

https://github.com/NethermindEth/near-sffl/blob/f63d4204156933e8e4bb23b3f4b832ec17969e73/plugin/cmd/main.go#L187-L188

We would like to add this feature to the `plugin` CLI. This requires extending the `AvsManager` (see https://github.com/NethermindEth/near-sffl/blob/main/operator/avs_manager.go)

As part of this feature we would also like to include proper tests that verify the expected behavior (`opt-in -> opt out`)

PR #171, opened 2024-05-20
Operator `opt-out`

emlautarom1:
Implements #170


emlautarom1:
I need to explore how to test this code first so marked as a draft for now.

operator/avs_manager.go
```diff
@@ -285,6 +286,21 @@ func (avsManager *AvsManager) RegisterOperatorWithAvs(
...
 }
 
+func (avsManager *AvsManager) DeregisterOperator(blsKeyPair *bls.KeyPair,) (error) {
+	// TODO: 'QuorumNums' is hardcoded for now
+	quorumNumbers := eigentypes.QuorumNums{0}
+	pubKey := eigenutils.ConvertToBN254G1Point(blsKeyPair.GetPubKeyG1())
+
+	_, err := avsManager.avsWriter.DeregisterOperator(context.Background(), quorumNumbers, pubKey)
+	if err != nil {
+		avsManager.logger.Errorf("Unable to deregister operator with avs registry coordinator")
```
taco-paco, line 296: Would be good to print error as well

operator/avs_manager.go
```diff
@@ -285,6 +286,21 @@ func (avsManager *AvsManager) RegisterOperatorWithAvs(
 	return nil
 }
 
+func (avsManager *AvsManager) DeregisterOperator(blsKeyPair *bls.KeyPair,) (error) {
```
taco-paco, line 289:
```suggestion
func (avsManager *AvsManager) DeregisterOperator(blsKeyPair *bls.KeyPair) error {
```

plugin/cmd/main.go
```diff
@@ -185,7 +183,15 @@ func plugin(ctx *cli.Context) {
...
-		fmt.Println("Opting out of slashing - unimplemented")
+		blsKeyPassword := ctx.GlobalString(BlsKeyPasswordFlag.Name)
+
+		blsKeypair, err := bls.ReadPrivateKeyFromFile(avsConfig.BlsPrivateKeyStorePath, blsKeyPassword)
+		if err != nil {
+			logger.Error("Failed to read bls private key", "err", err)
+			return
+		}
+
+		avsManager.DeregisterOperator(blsKeypair)
```
taco-paco, line 194: Error is never propagated

taco-paco (approved):
Good job!
One thing that I noticed is that `plugin` function doesn't return error so our app always exists gracefully with code 0 even if error occurred. 

Notice [here](https://github.com/urfave/cli/blob/1c29f69445b36de39dd2565c80d5328e6553354f/app.go#L525) we use deprecated signature. I would propose to change this

operator/avs_manager.go
emlautarom1: Surprising that `gofmt` does not format this automatically

plugin/cmd/main.go
emlautarom1: Added this feature: we now propagate all error values from `plugin`

operator/avs_manager.go
emlautarom1: Now we do: we propagate the error to `main.go` (see https://github.com/NethermindEth/near-sffl/pull/171/files#diff-b8b904d92cea98f8256ea3b62d416f66151d1315c5b58d870bfb553877724be4R201) and log the error in case there is any

emlautarom1:
After spending quite some time trying to test the code I'm struggling to come up with a solution. My intent is to write a test that registers an operator, checks that it's registered,  unregisteres it and checks that it's no longer registered.

Our entry point is `plugin/cmd/main.go`:

https://github.com/NethermindEth/near-sffl/blob/47e84e0fe342e19d2b7a3cfcbb4189d9fd13e3a0/plugin/cmd/main.go#L190-L203

There is no point in testing the CLI since it's responsible for reading the config values and passing them to `AvsManager` (the component that actually performs the logic), and passing back the resulting values. Thus, we're interested in testing `AvsManager`

`operator/avs_manager.go`

https://github.com/NethermindEth/near-sffl/blob/47e84e0fe342e19d2b7a3cfcbb4189d9fd13e3a0/operator/avs_manager.go#L289-L302

Here is the actual logic that we want to test, but in order to do so we need to construct an `AvsManager`. This struct is opaque, only exposing a `new` method that uses real dependencies (ex. `chainio.BuildAvsWriter`):

https://github.com/NethermindEth/near-sffl/blob/47e84e0fe342e19d2b7a3cfcbb4189d9fd13e3a0/operator/avs_manager.go#L62

We could add an additional constructor that allows us to pass the dependencies or expose the `AvsManager` struct fields but then we would need to mock things like `chainio.AvsWriter` which is the one doing the real work

https://github.com/NethermindEth/near-sffl/blob/47e84e0fe342e19d2b7a3cfcbb4189d9fd13e3a0/operator/avs_manager.go#L294

If we mock it, then the only thing we can assert is that we're calling the `DeregisterOperator` method, which is useless information.

---

At this point, I don't see any reason to continue pushing for a unit test: to test this behavior an integration test would be more reasonable since all the code is eventually coupled to an on-chain interaction with the AVS contracts.

I'd appreciate any suggestions on how this topic: @taco-paco @Hyodar

Hyodar:
I agree, this is better fitted for an integration test. Actually, we don't even use the plugin `deposit`, so we could remove it, and in the end this just initializes and wraps an `AvsManager`. That said, I don't think it's too relevant having a unit test for this.

We could still do it if it's desirable somehow, like:
1. We define an `OperatorPlugin` struct and turn our actions into receivers:
```go
type OperatorPlugin struct {
    blsKeyPair *bls.KeyPair
    ecdsaPrivateKey *ecdsa.PrivateKey
    avsManager AvsManagerer
}

func NewOperatorPlugin(/* same fields as struct */) *OperatorPlugin {
    // just create struct
}

func (*OperatorPlugin) OptOut() error {
    // ...
}
```
2. We generate a gomock from `AvsManagerer`
3. We use the mock in a test and expect calls to be made like that
```go
func createMockOperatorPlugin(mockCtrl *gomock.Controller, /* keypairs... */) (*OperatorPlugin, *mocks.MockAvsManagerer) {
    mockAvsManager := mocks.NewMockAvsManagerer(mockCtrl)
    return NewOperatorPlugin(/* keypairs... */, mockAvsManager), mockAvsManager
}

func SomeTest(t *testing.T) {
    mockCtrl := gomock.NewController(t)
    defer mockCtrl.Finish()

    mockOperatorPlugin, mockAvsManagerer := createMockOperatorPlugin(mockCtrl, /* ... */)
    
    mockAvsManagerer.EXPECT().DeregisterOperator(blsKeypair).Return(nil)
    err := mockOperatorPlugin.OptOut()
    assert.Nil(t, err)
}
```

Then, having unit tested `AvsManager`, we are assuming relaying the call is enough. You could also implement a more complete mock for `AvsManager` there that tracks the registrations and uses it for checking if an operator is registered.
Again, at this point I don't think it makes sense, and I'd agree in this case we care about the side effects that just mocking may not be enough, just a quick sketch showing one way to do it.

emlautarom1:
I ended up separating the `OperatorPlugin` into an interface with the three operations ("opt-in", "opt-out", "deposit") and a concrete implementation that uses the CLI to construct all dependencies (ex. `AvsManager`). This would allow us to construct an `OperatorPlugin` with explicit dependencies but testing is still hard: some method like `OptOut` are just getting data from the environment/filesystem and passing them to the `AvsManager`.

> Again, at this point I don't think it makes sense, and I'd agree in this case we care about the side effects that just mocking may not be enough, just a quick sketch showing one way to do it.

I agree and I don't think its worth pursuing any further with this task. I'll mark it as complete and we can revisit it if we find issues.

plugin/cmd/operator_plugin.go
```diff
@@ -0,0 +1,243 @@
...
+	_, err = o.clients.ElChainWriter.DepositERC20IntoStrategy(context.Background(), strategyAddr, amount)
+	if err != nil {
+		o.logger.Error("Failed to deposit into strategy", "err", err)
+		return err
+	}
+
+	return nil
+}
+
+func NewOperatorPluginFromCLIContext(ctx *cli.Context) (*CliOperatorPlugin, error) {
```
Hyodar, line 139: Just a matter of personal preference here, but wdyt about moving this up? Seems a bit more straightforward since this is a constructor.

plugin/cmd/operator_plugin.go
Hyodar, line 1: Ideally I'd suggest moving the base declaration to `plugin/plugin.go` or such and the cmd-specific to `/plugin/cmd`. But not a problem in this case, this can be work for when it's actually useful.

operator/avs_manager.go
```diff
@@ -285,6 +288,21 @@ func (avsManager *AvsManager) RegisterOperatorWithAvs(
...
+	// TODO: 'QuorumNums' is hardcoded for now
+	quorumNumbers := eigentypes.QuorumNums{0}
+	pubKey := eigenutils.ConvertToBN254G1Point(blsKeyPair.GetPubKeyG1())
+
+	_, err := avsManager.avsWriter.DeregisterOperator(context.Background(), quorumNumbers, pubKey)
+	if err != nil {
+		avsManager.logger.Errorf("Unable to deregister operator with avs registry coordinator: %v", err)
+		return err
+	}
+	avsManager.logger.Infof("Deregistered operator with avs registry coordinator.")
```
Hyodar, line 301: Could be a `Error` (using a tag for the error, like `logger.Error("msg", "err", err)`, which is how we do it in most places) and an `Info` - I see this happens in other places in `AvsManager`, not specific from this PR.

Hyodar (approved):

Hyodar:
Agreed. Only a few comments, LGTM.

plugin/cmd/operator_plugin.go
emlautarom1: I'll keep that in mind but for now I'll keep the current folder structure to not break the Dockerfile (https://github.com/NethermindEth/near-sffl/blob/9d45fbe61c0d057fd642323399cc6506839fdf63/plugin/cmd/Dockerfile)

PR #172, opened 2024-05-21
Fix double decrement on SafeEthClient wg

Hyodar:
Just noticed something that passed us by during #154 - `handleReinitEvent` has a `defer c.wg.Done()` even though the wait group is not incremented when it's called. It was probably a byproduct of a refactor, as now it spawns it's own task. In this case, it makes sense to just remove it.


taco-paco:
Yeah, I missed this as well. Makes sense

taco-paco (approved):

PR #173, opened 2024-05-22
Fix likely deadlock issue in SafeClient

Hyodar:
From our testing data on operator state root messaging, it seems quite likely there's a deadlock happening in `SafeClient` for block submissions. Would be good to get an overall review on it. Trying to change some of the approaches that seem problematic while I get more info on it, at the moment:
* Centralized resubbing in reinits, so there's no way a resub can happen while it's reinitializing
* Made resubs not block the event listening from the subscription task

Good to highlight the rollup clients are only used for header subscriptions, so it's down to `SubscribeNewHead` specifically.

### DIFF
--- a/core/safeclient/client.go
+++ b/core/safeclient/client.go
@@ -505,19 +505,13 @@ func (c *SafeEthClient) SubscribeNewHead(ctx context.Context, ch chan<- *types.H
 				}
 			case <-sub.Err():
 				c.logger.Info("Underlying subscription to new heads ended, resubscribing")
-				err := resub()
-				if err = c.handleClientError(err); err != nil {
-					c.logger.Error("Failed to resubscribe to new heads", "err", err)
-				}
+				c.triggerReinit()
 			case <-headerTicker.C:
 				c.logger.Info("Header ticker fired, ending subscription")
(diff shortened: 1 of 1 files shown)


Hyodar:
One additional point here is - is it necessary to reinitialize the client? It makes sense to do it, but maybe it's not necessary. I'm starting to feel like it's not, and we could just simplify this module a lot.

emlautarom1:
How was this deadlock detected? Any way to reproduce in a testing environment?

Hyodar:
As mentioned there, not totally sure what's happening yet. Couldn't reproduce it as well, currently trying to subscribe to headers locally and see if I can catch it. The symptoms are simply that randomly after some time an operator stops sending `StateRootUpdateMessage`s, also trying to get more info on it from operator logs.

Hyodar:
Setting this as draft for now, as I couldn't yet reproduce the issue and I didn't see it happening again. Also, it's quite likely I'm going to simplify the module a bit - we don't actually need to reinitialize the client.

Hyodar:
Closing as this is not relevant anymore as per latest changes.

PR #177, opened 2024-05-23
Add operator initialization metric to aggregator

Hyodar:
Adds a simple metric to track operator initializations in aggregator. The idea here is not that we strictly need to know that operator X has initialized, but at least have some kind of ping so we can detect crashes if they happen.

### DIFF
--- a/aggregator/metrics.go
+++ b/aggregator/metrics.go
@@ -14,7 +14,7 @@ const (
 )
 
 type RpcEventListener interface {
-	IncOperatorReinitializations(operatorId [32]byte)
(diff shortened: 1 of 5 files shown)


emlautarom1 (approved):
LGTM

I don't see anything worth mentioning except for the usage of `logger.Errof` instead of `logger.Error` with tags (I'm not sure if there's an agreement on what to use)

Hyodar:
Oh well. Yeah, we've not settled on one specifically, but we've been trying to use `Error` + tag `"err"`. I see there are multiple places in metrics that the format one is used, so I don't think this is work for this PR. Added an issue for that: #179.

PR #178, opened 2024-05-23
Add plugin-specific setup sub-directory

Hyodar:
It's not that nice having the plugin together with the operator `setup`. The most notable situation is if an operator doesn't want to use the `RegisterOnStartup` setting to preserve their private key. In this case, it's not practical to change the config or have two copies of the repo.
This PR adds its own `setup/` sub-directory to the operator plugin, with a separate config file, `.env` and so on.


Hyodar:
I think the only missing change is docs-related. Going to do that soon, but opening for reviews already

setup/plugin/config/operator.yaml
```diff
@@ -0,0 +1,15 @@
...
+avs_registry_coordinator_address: 0x0069A298e68c09B047E5447b3b762E42114a99a2
+operator_state_retriever_address: 0x8D0b27Df027bc5C41855Da352Da4B5B2C406c1F0
+
+# AVS network RPCs
+eth_rpc_url: https://ethereum-holesky-rpc.publicnode.com
+eth_ws_url: wss://ethereum-holesky-rpc.publicnode.com
+
+# EigenLayer ECDSA and BLS private key paths
+ecdsa_private_key_store_path: /near-sffl/config/keys/ecdsa.json
+bls_private_key_store_path: /near-sffl/config/keys/bls.json
```
emlautarom1, line 15: Not sure about this one, but shouldn't these paths be relative (`./`) rather than absolute (`/`)?

emlautarom1 (approved):
LGTM

setup/operator/config/operator.yaml
```diff
@@ -32,8 +32,8 @@ node_api_ip_port_address: 0.0.0.0:9010
 # Whether to try and register the operator in the AVS and in EL on startup.
 # It will not re-register the operator if already registered.
 # If unset, the operator will not be registered on startup! You'll need to
-# manually register the operator - not recommended
-register_operator_on_startup: true
+# manually register the operator.
+register_operator_on_startup: false
```
taco-paco, line 36: Is that parameter planned to be used - ever to be true? It is quite an implicit one. Would be nice to get rid of it.

taco-paco (approved):

setup/operator/config/operator.yaml
Hyodar: I'd like to keep it, as it's convenient for a user that might want to use it and also for testing

PR #183, opened 2024-05-24
Standardize subscription error handling

Hyodar:
Now that all our subscriptions are made through a `SafeEthClient`, the only situations in which there is a subscription error are either when the client is being closed or the context is ended - both cases mean deliberately exiting (and in our case program exit). So, we can just standardize unsubbing and (if receiver) closing the channel whenever there is an error.
Ideally none of these steps would be necessary and are functionally changing anything if we assume the unsubbing has already happened and the channel can be closed automatically on GC, but if there ever is some extra reason at least we're not leaking anything.


emlautarom1 (approved):
LGTM, I assume that there are no issues with the test suite

PR #185, opened 2024-05-27
Fix plugin registration and add registration int test

Hyodar:
Something that passed us by on #171 - the plugin receivers considered the context would be part of the struct, but it wasn't being assigned on the constructor. For now, I removed the interface, as we're not having much use for it, and made it so the receivers get the context. Saving contexts is a bit weird, so I think this is for the best, and we can get a common interface later.
I also added an integration test for opt-in/opt-out operations.

It's important to merge this quickly so we have registering working.

### DIFF
--- a/plugin/cmd/operator_plugin.go
+++ b/plugin/cmd/operator_plugin.go
@@ -57,7 +57,7 @@ func NewOperatorPluginFromCLIContext(ctx *cli.Context) (*CliOperatorPlugin, erro
 		RegistryCoordinatorAddr:    avsConfig.AVSRegistryCoordinatorAddress,
 		OperatorStateRetrieverAddr: avsConfig.OperatorStateRetrieverAddress,
 		AvsName:                    "super-fast-finality-layer",
-		PromMetricsIpPortAddress:   avsConfig.EigenMetricsIpPortAddress,
+		PromMetricsIpPortAddress:   "127.0.0.1:9090",
 	}
 
 	ethHttpClient, err := eth.NewClient(avsConfig.EthRpcUrl)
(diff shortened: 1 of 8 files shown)


plugin/cmd/operator_plugin.go
```diff
@@ -66,7 +57,7 @@ func NewOperatorPluginFromCLIContext(ctx *cli.Context) (*CliOperatorPlugin, erro
 		RegistryCoordinatorAddr:    avsConfig.AVSRegistryCoordinatorAddress,
 		OperatorStateRetrieverAddr: avsConfig.OperatorStateRetrieverAddress,
 		AvsName:                    "super-fast-finality-layer",
-		PromMetricsIpPortAddress:   avsConfig.EigenMetricsIpPortAddress,
+		PromMetricsIpPortAddress:   "127.0.0.1:9090",
```
emlautarom1, line 60: Why are we hard-coding this value now?

emlautarom1 (approved):
LGTM.

I need to spend some time looking into test containers (this was the kind of test I intended to have in #171)

emlautarom1:
On the operator-plugin side then the only relevant change is that we no longer store the `cli.Context` but prefer to pass it as argument when needed. Is this a common Go practice (not storing context?)

plugin/cmd/operator_plugin.go
Hyodar: Because it's not actually necessary - check #178. It's basically a limitation on using some eigensdk utilities.

Hyodar:
Exactly - storing contexts is not the norm.

PR #187, opened 2024-05-27
Remove header ticker stopping on SafeEthClient

Hyodar:
We were stopping the header ticker once a successful resub happened, but we actually *should* leave it running because sometimes providers just stop sending blocks without an underlying error. With this, we do try to resub if there are no blocks received in that time window.


emlautarom1 (approved):
:+1:

PR #188, opened 2024-05-27
Fix operatorId not being assigned to AggregatorRpcClient

Hyodar:


emlautarom1 (approved):
:+1:

This is not the first time we had issues with 0-initialized values (see #185) so it would be worth to investigate some sort of static analyzer

PR #189, opened 2024-05-27
Capture `err` for proper check

emlautarom1:
There are certain cases where we are not properly setting the usual `err` variable to the latest error, making checks (`if err != nil`) useless since they're double-checking the same error.


Hyodar (approved):
LGTM

PR #192, opened 2024-05-27
Fix RMQ recover condition check

Hyodar:
This condition is not easily reached at all, but a low-hanging fix.


emlautarom1 (approved):
:+1:

PR #194, opened 2024-05-28
feat: custom pubkeys inmemory service

taco-paco:


aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+
+		ors.logger.Debug("Subscribing to new pubkey registration events on blsApkRegistry contract", "service", "OperatorRegistrationsServiceInMemory")
+		newPubkeyRegistrationC, newPubkeyRegistrationSub, err := ors.avsRegistrySubscriber.SubscribeToNewPubkeyRegistrations()
+		if err != nil {
+			ors.logger.Error("Fatal error opening websocket subscription for new pubkey registrations", "err", err, "service", "OperatorRegistrationsServiceInMemory")
+			// see the warning above the struct definition to understand why we panic here
+			panic(err)
+		}
+
+		ors.queryPastRegisteredOperatorEventsAndFillDb(ctx, pubkeyByAddrDict, pubkeyByIdDict)
```
emlautarom1, line 93: Could we lift this block outside the goroutine, leaving only the `for` in it? I think we could avoid the `WaitGroup` at all which makes things a bit confusing

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+
+			case err := <-newPubkeyRegistrationSub.Err():
+				ors.logger.Error("Error in websocket subscription for new pubkey registration events. Attempting to reconnect...", "err", err, "service", "OperatorRegistrationsServiceInMemory")
+				newPubkeyRegistrationSub.Unsubscribe()
+				newPubkeyRegistrationC, newPubkeyRegistrationSub, err = ors.avsRegistrySubscriber.SubscribeToNewPubkeyRegistrations()
+				if err != nil {
+					ors.logger.Error("Error opening websocket subscription for new pubkey registrations", "err", err, "service", "OperatorRegistrationsServiceInMemory")
+					// see the warning above the struct definition to understand why we panic here
+					panic(err)
+				}
```
emlautarom1, line 113: Do we also want to `panic` on reconnects? Does not seem to be the same case as during construction (maybe a copy/paste issue?)

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	ors.logger.Debug("List of queried operator registration events in blsApkRegistry", "alreadyRegisteredOperatorAddr", alreadyRegisteredOperatorAddrs, "service", "OperatorRegistrationsServiceInMemory")
+	for i, operatorAddr := range alreadyRegisteredOperatorAddrs {
+		operatorPubkeys := alreadyRegisteredOperatorPubkeys[i]
+		pubkeyByAddrDict[operatorAddr] = operatorPubkeys
+
+		operatorId := types.OperatorIdFromPubkey(operatorPubkeys.G1Pubkey)
+		pubkeyByIdDict[operatorId] = operatorPubkeys
+	}
```
emlautarom1, line 165: I would just inline this code, it provides no value to the reader to have it separated from the only call site

emlautarom1 (changes requested):
I would appreciate some context on why this feature is necessary. The branch name suggests it has something to do with DoS but it's not clear from the code

Hyodar:
Jumping in the conversation - still need to review it though, most likely not going to be able to do it today. The thing here is currently the aggregator makes a bunch of RPC calls to initialize a message even though the signature present is not from a registered operator, which opens a quite big DOS window. So, the idea is one of the most basic mitigations is we validate the signature before anything else - and the OperatorPubkeys service from `eigensdk` is not flexible enough.

aggregator/operator_registrations_inmemory.go
taco-paco: `WaitGroup::Wait` parks calling goroutine, allowing scheduler to execute another one until WaitGroup's counter is 0. Proposed implementation changes asynchronous code to synchronous where the thread will hang until `ors.queryPastRegisteredOperatorEventsAndFillDb` finishes and won't be able to do smth else.

aggregator/operator_registrations_inmemory.go
taco-paco: You mean inline this function in `startServiceInGoroutine`? It provides value by simplifying reader's flow in main logic block within startServiceInGoroutine by slicing it into execution of smaller logic blocks, so one doesn't have to read the whole thing in one function.

aggregator/operator_registrations_inmemory.go
taco-paco: See `runtime_Semacquire`

emlautarom1: Seems mostly like a runtime difference on which thread the `ors.queryPastRegisteredOperatorEventsAndFillDb` will be executed. In both cases the caller of the constructor `NewOperatorRegistrationsServiceInMemory` will be blocked due to the `wg.Wait`

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+		ors.queryPastRegisteredOperatorEventsAndFillDb(ctx, pubkeyByAddrDict, pubkeyByIdDict)
+
+		// The constructor can return after we have backfilled the db by querying the events of operators that have registered with the blsApkRegistry
+		// before the block at which we started the ws subscription above
+		wg.Done()
+
+		for {
+			select {
+			case <-ctx.Done():
+				ors.logger.Infof("OperatorRegistrationsServiceInMemory: Context cancelled, exiting")
```
Hyodar, line 102: Unnecessary format here

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+var _ operatorpubkeys.OperatorPubkeysService = (*OperatorRegistrationsServiceInMemory)(nil)
+
+// NewOperatorRegistrationsServiceInMemory constructs a OperatorRegistrationsServiceInMemory and starts it in a goroutine.
+// It takes a context as argument because the "backfilling" of the database is done inside this constructor,
+// so we wait for all past NewPubkeyRegistration events to be queried and the db to be filled before returning the service.
+// The constructor is thus following a RAII-like pattern, of initializing the serving during construction.
+// Using a separate initialize() function might lead to some users forgetting to call it and the service not behaving properly.
+func NewOperatorRegistrationsServiceInMemory(
+	ctx context.Context,
+	avsRegistrySubscriber avsregistry.AvsRegistrySubscriber,
```
Hyodar, line 55: IMO we should send a SafeClient and the BLSRegistry address here, create the bindings in our own and manage them here. Reason for that is it's just simpler, we don't seem to have a good reason to use a separate module. Or we should at least make sure we are using a SafeClient when instantiating this so we can rely on it for resubs and such.

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+			ors.logger.Error("Fatal error opening websocket subscription for new pubkey registrations", "err", err, "service", "OperatorRegistrationsServiceInMemory")
+			// see the warning above the struct definition to understand why we panic here
+			panic(err)
+		}
+
+		ors.queryPastRegisteredOperatorEventsAndFillDb(ctx, pubkeyByAddrDict, pubkeyByIdDict)
+
+		// The constructor can return after we have backfilled the db by querying the events of operators that have registered with the blsApkRegistry
+		// before the block at which we started the ws subscription above
+		wg.Done()
```
Hyodar, line 97: What is the rationale for handling the dict in the service? Overall it seems better to just initialize it directly in the constructor and trigger the 'task' later, it's already waiting anyway. Would avoid the panic as well.

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+
+		for {
+			select {
+			case <-ctx.Done():
+				ors.logger.Infof("OperatorRegistrationsServiceInMemory: Context cancelled, exiting")
+				return
+
+			case err := <-newPubkeyRegistrationSub.Err():
+				ors.logger.Error("Error in websocket subscription for new pubkey registration events. Attempting to reconnect...", "err", err, "service", "OperatorRegistrationsServiceInMemory")
```
Hyodar, line 106: We can just give up here if it's a SafeClient, no need to worry about resubs

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+				operatorPubkeyQuery.respC <- resp{pubkeys, ok}
+			}
+		}
+	}()
+}
+
+func (ors *OperatorRegistrationsServiceInMemory) queryPastRegisteredOperatorEventsAndFillDb(ctx context.Context, pubkeyByAddrDict map[common.Address]types.OperatorPubkeys, pubkeyByIdDict map[types.OperatorId]types.OperatorPubkeys) {
+	// Querying with nil startBlock and stopBlock will return all events. It doesn't matter if we queryByAddr some events that we will receive again in the websocket,
+	// since we will just overwrite the pubkey dict with the same values.
+	alreadyRegisteredOperatorAddrs, alreadyRegisteredOperatorPubkeys, err := ors.avsRegistryReader.QueryExistingRegisteredOperatorPubKeys(ctx, nil, nil)
```
Hyodar, line 152: Shouldn't we chunk blocks here? i.e. 10000 blocks / call.

Hyodar: Another approach would be going with nil, nil first and chunks only if errored

aggregator/operator_registrations_inmemory.go
```diff
@@ -0,0 +1,190 @@
...
+	// channel through which to receive the response (operator pubkeys)
+	respC chan<- resp
+}
+type queryById struct {
+	operatorId types.OperatorId
+	// channel through which to receive the response (operator pubkeys)
+	respC chan<- resp
+}
+
+type resp struct {
```
Hyodar, line 40: This is probably from eigensdk itself, but maybe `queryResp` would already be an improvement

aggregator/rpc_server.go
```diff
@@ -40,25 +41,50 @@ func (agg *Aggregator) startServer() error {
...
 		agg.logger.Error("Failed to get task response digest", "err", err)
 		return TaskResponseDigestNotFoundError500
 	}
 
+	operatorId := signedCheckpointTaskResponse.OperatorId
+	signature := signedCheckpointTaskResponse.BlsSignature
+	ok, err := agg.verifyOperatorIdRegistered(operatorId, taskResponseDigest, signature)
+	if err != nil {
+		agg.logger.Error("Failed to verify message", "err", err)
+		return err
```
Hyodar, line 74: IMO we should either add a new error or just return it as SignatureVerificationFailed anyway. Wdyt? Applies to all similar cases

aggregator/rpc_server_test.go
```diff
@@ -73,15 +76,42 @@ func TestProcessSignedStateRootUpdateMessage(t *testing.T) {
...
+		BlockHeight:         2,
+		Timestamp:           3,
+		NearDaCommitment:    keccak256(4),
+		NearDaTransactionId: keccak256(5),
+		StateRoot:           keccak256(6),
+	}
+
+	signedMessage, err := createMockSignedStateRootUpdateMessage(message, *MOCK_OPERATOR_KEYPAIR)
+	assert.Nil(t, err)
+	invalidateSignature(&signedMessage.BlsSignature)
```
Hyodar, line 103: Could this just reassign the signature to 0? Wasn't totally clear to me when reading what invalidating a signature meant

Hyodar (changes requested):
Overall LGTM, just still one or two changes I think are considerable enough to require a quick look again.

aggregator/operator_registrations_inmemory.go
taco-paco: https://github.com/NethermindEth/near-sffl/pull/194#discussion_r1620326046

I agree with initial implementation here. Right now constructor look like
`async func NewOperatorRegistrationsServiceInMemory.` What you proposing is 
`func NewOperatorRegistrationsServiceInMemory`. 

In first case you delegate task to scheduler in second you say: I will filter all events right now disregarding all pending tasks in queue.  As developer calling a constructor that does intensive computation is an unexpected surprise. It will do it later anyway but when - is scheduler business.

Imagine a React.JS where while a scheduled task is rerender you start queuing whole history blocking it.

aggregator/operator_registrations_inmemory.go
taco-paco: > Shouldn't we chunk blocks here? i.e. 10000 blocks / call.

This is how its done

aggregator/operator_registrations_inmemory.go
Hyodar: I agree in a sense, but the constructor is waiting for it anyway

Hyodar: In the end it's like the discussion outcome in the linked comment - for the caller it doesn't really change the situation much, the work is being done in another goroutine but it's being waited anyway. IMO if it's being waited I don't think we have much of a use for doing it in a goroutine, especially with the error handling considerations

aggregator/rpc_server_test.go
Hyodar: Yeah, exactly. It just feels like it's better to try with a signature we know is invalid as opposed to calling something - doing it like that just feels like it depends on some side effect

taco-paco: Not sure what you mean its better. Negating valid signature - produces invalid one.

aggregator/operator_registrations_inmemory.go
taco-paco: > I agree in a sense, but the constructor is waiting for it anyway

Exactly, caller is parked and that enables scheduler to progress with other tasks: such as render(in React) or any other scheduled tasks. Intensive work shall be scheduled.

>  for the caller it doesn't really change situation much

It's not only about the current caller. If that wouldn't change anything such things as schedulers in OSs, languages wouldn't exist and everything could just be executed on the thread without asynchronous tasks. 

Imagine a single process running blocking keyboard inputs/SSD and other drivers signals, because it doesn't want to let other tasks to be scheduled.

Hyodar: Alright, it seemed to me you hadn't noticed the `wg.Done()` somehow. I'm not disagreeing with the approach specifically, just IMO we still don't really have a use for this extra step in the context we're using this module. If we are doing it like that though, then we should not just throw panics around and we should also do it the 'right way' and e.g. have an error channel or such that is read in the constructor to relay the error.

aggregator/operator_registrations_inmemory.go
Hyodar: Btw, maybe we can add our custom impl to this as well. Calling in chunks should take way more time than an unbound block range call, so I'd actually prefer it (and should save us quite some time) if we tried unbound first before moving to chunks.

aggregator/rest_server_test.go
```diff
@@ -18,7 +18,7 @@ func TestGetStateRootUpdateAggregation(t *testing.T) {
 	mockCtrl := gomock.NewController(t)
 	defer mockCtrl.Finish()
 
-	aggregator, _, _, _, _, _, mockDb, _, _, err := createMockAggregator(mockCtrl, MOCK_OPERATOR_PUBKEY_DICT)
+	aggregator, _, _, _, _, _, _, mockDb, _, _, err := createMockAggregator(mockCtrl, MOCK_OPERATOR_PUBKEY_DICT)
```
Hyodar, line 21: Small comment not for this PR - seems like it's time to export these returns in a neater way. Before we had only a few returns, but now it's quite a few mocks and `_, `, we should just organize this in a struct.

aggregator/operator_registrations_inmemory.go
taco-paco: As I remember that was a frequent issue, so I'm not sure we shall be concerned with this.

Do you remember request error after exceeding some time of execution? In that case it would be a problematic, since we would be blocked by unbound request for a while and then receive error that it "exceeded time limit and doesn't work anymore"

Hyodar: No, at least the aggregator should use a node that supports that - currently we are. I'd say even not knowing the RPC it's still worth it. Making calls in chunks overall increases the call time considerably, it's just a matter of following block range limits of RPCs.

aggregator/operator_registrations_inmemory.go
taco-paco: Fixed

aggregator/operator_registrations_inmemory.go
taco-paco: Removed

aggregator/operator_registrations_inmemory.go
taco-paco: Done in a slightly different way:
Using our AvsSubscriber instead which uses SafeClient
AvsSubscriber now implements AvsRegistrySubscriber interface to be able to work with registry

aggregator/operator_registrations_inmemory.go
taco-paco: Giving up now, though the service still works responds unitil ctx.Done

aggregator/rpc_server.go
emlautarom1: Agree mostly for consistency

aggregator/rpc_server_test.go
emlautarom1: I aggree with the comment but I would do it differently:

```suggestion
	signedMessage.BlsSignature = NewInvalidSingature()
```

Passing a field by reference to mutate it is what makes it confusing. Either way is just an opinion

aggregator/operator_registrations_inmemory.go
emlautarom1: I see that we're using the eigensdk `AvsRegistrySubscriber` rather than our own `AvsSubscriber`. Is the comment outdated?

emlautarom1 (approved):
LGTM

aggregator/operator_registrations_inmemory.go
taco-paco: Not really. AvsSubscriber implements methods of AvsRegistrySubscriber. 

From`aggregator` we pass AvsSubscriber into NewOperatorRegistrationsServiceInMemory. Since `interface AvsSubscriber: AvsRegistrySubscriber` we can and use `AvsRegistrySubscriber` here directly since we need only its methods.

taco-paco: This enables use of this struct with both our AvsSubscriber and el AvsRegistrySubscriber, in case one wants to

aggregator/operator_registrations_inmemory.go
Hyodar: With the way you were describing before, I wasn't sure you were aware it was waiting for the async task anyway. That said, if you're removing the panics and catching the error, that's fine by me

Hyodar: Do we need wg considering we are waiting for the error channel anyway?

Hyodar:
Can we merge this soon? @emlautarom1 @taco-paco

emlautarom1:
LGTM. Just need to solve the merge conflicts + resolve discussions.

aggregator/operator_registrations_inmemory.go
taco-paco: Would be good to merge it in the current state and add required logic in separate PR

aggregator/operator_registrations_inmemory.go
Hyodar: Keeping it like this for consistency with `eigen-sdk`

aggregator/rpc_server.go
Hyodar: I now moved this check to `Aggregator` directly, and the error is mapped inside `RpcServer`

aggregator/rest_server_test.go
Hyodar: Added as an issue in #270

aggregator/rpc_server_test.go
Hyodar: Done in [bc67968](https://github.com/NethermindEth/near-sffl/pull/194/commits/bc679685e81a3f20ebf09d443395b20dfb5def13)

PR #195, opened 2024-05-28
Add forceOperatorSetUpdate to SFFLRegistryRollup

Hyodar:
This PR adds a testnet-specific utility for managing unattested operator set updates. Currently, as we don't have a reliable delivery for operator set updates, there can be a situation in which operators don't agree on an operator set update for inactivity. Especially in an unstable testing environment this is a very much possible situation.
In this case, we don't have an agreement to send to the rollup contracts. One solution for this, rollup-specific, would be accepting bridged calls and using the rollup bridge to do that. However, for now having an owner method for forcing an operator set update is a lot more convenient and doesn't really change the trust assumption, as e.g. the admin could still change the quorum threshold or update the contracts in order to achieve the same result. So, in the end, this is just a shortcut so we don't have to make a quick workaround if this ever is needed. In a mainnet scenario, this method would be either non-existent or managed by a fancier flow, such as by using a timelock or governance admin.

After approval, the rollup contracts will be upgraded to use this new implementation. The understanding is this should be used with considerable care only, and the situation which caused the consensus to not be reached (or some aggregator fault) should be investigated.


contracts/evm/src/rollup/SFFLRegistryRollup.sol
```diff
@@ -145,6 +145,21 @@ contract SFFLRegistryRollup is Initializable, OwnableUpgradeable, Pausable, SFFL
...
 
+    /**
+     * @notice Forces an operator set update. This is meant to be used only
+     * by the owner in a testnet scenario in case there is no consensus on a
+     * particular operator set update. This can also be used while operator
+     * set updating is paused.
+     * @param message Operator set update message
+     */
+    function forceOperatorSetUpdate(OperatorSetUpdate.Message calldata message) external onlyOwner {
+        require(message.id == nextOperatorUpdateId, "Wrong message ID");
```
emlautarom1, line 156: There is no test that triggers this `require`

emlautarom1 (approved):
It would be nice for this PR to update the generated bindings:

https://github.com/NethermindEth/near-sffl/blob/63b7376c5ace5b3fd423d564779bb45047bfbb3d/contracts/bindings/SFFLRegistryCoordinator/binding.go#L1

Hyodar:
Sure, going to update the bindings and devnet anvil states soon. Also going to replicate the `updateOperatorSet` tests for this one.

contracts/evm/src/rollup/SFFLRegistryRollup.sol
Hyodar: Done in [61aaeaf](https://github.com/NethermindEth/near-sffl/pull/195/commits/61aaeafdae58cd6dd98631366990309177803f17)

PR #196, opened 2024-05-29
Improve resubscription error handling

Hyodar:
This PR makes two small improvements on resubscriptions:
1. If the subscription didn't fail but the log filtering did, the underlying subscription is still changed
2. The resub ticker is reset no matter what, as in the end we're rate limiting resubs


emlautarom1 (approved):
We need to start adding more localized tests for `safeclient`

Hyodar:
Agreed. In the end, these weren't total bugs, just not-necessarily-great ideas.

PR #200, opened 2024-05-29
Fix `SafeEthClient` concurrent close

### LINKED ISSUE #199: Unsafe `Close` in `SafeEthClient`
The current `SafeEthClient` is prone to race conditions:

https://github.com/NethermindEth/near-sffl/blob/4dabcd9eca5a1b3df98957913b93d26914b20c9c/core/safeclient/client.go#L303-L313

It's possible for two threads to run `CLose` simultaneously while `c.close == false`. Then, both threads will `close` the `c.closeC` channel triggering a panic due to double-closing.

emlautarom1:
Fixes #199 by using `sync.Once` to avoid double-closing a channel

### DIFF
--- a/core/safeclient/client.go
+++ b/core/safeclient/client.go
@@ -35,13 +35,13 @@ type SafeEthClient struct {
 	logger           logging.Logger
 	rpcUrl           string
 	closeC           chan struct{}
-	closed           bool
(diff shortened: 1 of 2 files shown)


Hyodar (approved):
LGTM!

PR #201, opened 2024-05-29
Only resub on underlying sub error

Hyodar:
Some provider nodes simply don't answer `unsubscribe`s timely, which makes this stall the log processing. This actually doesn't seem to be our fault - testing with self-hosted and local nodes works, but it is what it is. Instead, for now this makes it so we're only re-subbing logs on an actual underlying sub error and we can see how things behave.


taco-paco (approved):

emlautarom1 (approved):

Hyodar:
I'm putting this on hold for now. On our node, unsubscribing works as intended, so we may want to do some extra analysis here for other providers.

PR #203, opened 2024-05-31
Additional `SafeEthClient` tests

emlautarom1:
Some additional tests to `SafeEthClient` and `SafeSubscription`. They haven't uncovered any bugs yet but they're worth having in the case we want to refactor any of the tested modules


Hyodar (approved):
LGTM

PR #204, opened 2024-06-03
Separated RpcServer & RestServer from aggregator

taco-paco:
This closes: Specify Aggregator methods to be exposed in RPC. Done as result of below
And tech debt of separating rpc and rest servers from aggregator


aggregator/rpc_server/server.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	s.listener = listener
+	return nil
+}
+
+func (s *RpcServer) Start() error {
+	s.logger.Info("Starting aggregator rpc server.")
+
+	err := rpc.Register(s)
```
Hyodar, line 69: Checking this PR deeply later, but wanted to highlight something I noticed.

Here we are changing the name used for the RPC calls (can be seen in `rpc_client.go`). I'd still prefer it to be registered as `Aggregator`, as it's the aggregator RPC server indeed, but either way - keeping it like this is a **breaking change**. I think it wouldn't be if we kept the name.
That applies to all PRs from everyone: whenever what you're doing is a breaking change, make this clear somehow so it's properly handled, especially in the PR description. Reviewers should also be aware of this whenever reviewing, as it may pass unnoticed by the author. @taco-paco @emlautarom1 @mikhailUshakoff

taco-paco: Agree

aggregator/cmd/main.go
```diff
@@ -69,6 +71,24 @@ func aggregatorMain(ctx *cli.Context) error {
...
+	}
+
+	go rpcServer.Start()
+
+	restServer := restserver.NewRestServer(config.AggregatorRestServerIpPortAddr, agg, logger)
+	if registry != nil {
+		if err = restServer.EnableMetrics(registry); err != nil {
+			return err
+		}
+	}
```
Hyodar, line 89: Nit-pick: extra whitespace to match with the other one

aggregator/cmd/main.go
```diff
@@ -69,6 +71,24 @@ func aggregatorMain(ctx *cli.Context) error {
...
+	go restServer.Start()
```
Hyodar, line 90: I agree with separating the RPC and REST server as modules, but not sure if it makes sense totally separating them from the aggregator, just seems like extra boilerplate. Imo we should still have an `Aggregator::startRpcServer` and `Aggregator::startRestServer` that is called on the construtor or, if that's really the intention here, on the cmd main.

Hyodar (approved):
GJ, LGTM. The only consideration I have is in terms of separately declaring the RPC and REST clients, but let's keep the discussion of that in the comment.

aggregator/cmd/main.go
taco-paco: Not sure there's any boilerplate. Servers just propagate calls lower to an application but they shall be a sperate entity.

They need to handle various connection scenarios which are not related in any sense to `aggregator`. Those are out of the scope of this PR.

Right now it may look like it is some dummy server propagating calls lower but this PR only intents to separate them  and not add extra logic on top

taco-paco: As soon as we will try to introduce extra logic on top of servers: limiting requests rate, reconnections or anything else our aggregator will become a frankenstein handling all of that.

Hyodar: No, as I said, I agree with separating them as modules, I don't agree that much with not including them as part of the aggregator and having to create them separately, especially when we have to e.g. expose the prometheus registry from the aggregator in order to do that. I do think they should be separate, I just think the aggregator should have them inside of it and at least some option to create them on the constructor.

Hyodar: We don't really have a use-case that justifies this boilerplate above, we always want to expose the aggregator.

Hyodar: Well, we do have an issue with that in terms of packages though, since the naive approach would result in a cyclic dependency I think. I don't have an issue keeping the servers in the `aggregator` package though.

Hyodar: @emlautarom1 let us know your opinion on this later. Thinking more about it, I don't mind it too much, but I'd still like more opinions.

taco-paco: Aggregator shall not know anything about servers. That's IMO incorrect hierarchy. Aggregator is an app for servers. As you noticed having composition or simply creating servers within `Aggregator` will result in circular dependency which is a signal that architecture is broken. 

Having a package that will be only  creating/controlling servers&aggregator is an overkill.

> expose the prometheus registry from the aggregator in order to do that.

That is a hack, I agree but only because of in our eigensdk version API that creates registry itself instead of taking it as an argument.

taco-paco: I made a [PR](https://github.com/Layr-Labs/eigensdk-go/pull/117/files) a while ago to eigensdk to improve such behavior. The changes are merged in the later versions.

I guess I can mark this section as TODO for the future removal once we update eigensdk version

Hyodar: > Thought of having a package that will be only creating/controlling servers&aggregator but considered it an overkill.

That is my understanding, I agree with the package hierarchy, I just don't like that we have to explicitly expose it through instantiating separate clients when we know we're going to do it everytime. At the same time, with a separate package to handle both we get into a 'subpackage hell'. Would like some others' opinions on that.
My impression is that in go projects people are usually less strict on separating packages.

Hyodar: There's an easier solution for this. We can also put the common interfaces in `aggregator/types` - with that, the e.g. RPC package only knows the interface types and the aggregator can just instantiate it, either containing it or just having a utility method for that.

taco-paco: I'm not following why servers shall be started from `Aggregator`. 
Such approach basically says - when Aggregator instance is created, we start some servers. That is shall not be true at least for integration/unit tests of aggregator. 

Aggregator - aggregates signatures pretty much and does some db work. Its not clear for me why it shall start servers or anything.

taco-paco: Right now we don't have subpackage hell, and I don't think this will lead to this.

The solution that could be applied here, not that I like it. We move servers pack to aggregator package but keep them as separate structures. Create a separate function in utils for creating all of them and conclude on that.

But I don't like above since it is to break heirarchy since compiler won't hit us with circular dependency anymore.

Everything else is attempt to go around compiler in order to make it work. If we need such hacks that means our structure is broken

Hyodar: The ideal way would be to have an extra module that holds everything together, but we did agree on it being overkill. That said, separately initializing everything that isn't aggregating would mean separately initializing lots of components of aggregator. I don't think it makes sense it's necessary to add boilerplate for initializing servers all the time when it's something we always want to do - especially if we think about lifetime and such, throwing separate components around is just getting things more complicated than they need to be.

Hyodar: IMO making the bridge with aggregator types is the best solution we have here - the servers only know the aggregator interface they need to know and the aggregator can know the servers.

Hyodar: Not that I don't agree with the current approach in a way. I'd like more opinions on this - if this takes too long we can also just leave it the way it is and move on.

Hyodar: For now, though, one thing I'd ask you to do - since we're going to do the `GetRegistry` hack anyway, you could move the metrics enabling to the constructor itself, it gets considerably less ugly then.

aggregator/aggregator.go
```diff
@@ -523,3 +531,164 @@ func (agg *Aggregator) handleOperatorSetUpdateReachedQuorum(ctx context.Context,
...
+		return DigestError
+	}
+
+	err = agg.stateRootUpdateBlsAggregationService.InitializeMessageIfNotExists(
+		messageDigest,
+		coretypes.QUORUM_NUMBERS,
+		[]eigentypes.QuorumThresholdPercentage{types.MESSAGE_AGGREGATION_QUORUM_THRESHOLD},
+		types.MESSAGE_TTL,
+		types.MESSAGE_BLS_AGGREGATION_TIMEOUT,
+		0,
```
emlautarom1, line 577: Why is `ethBlockNumber = 0` here? Does it mean that it's irrelevant?

emlautarom1 (changes requested):

emlautarom1 (changes requested):
Some small details that could have been overlooked but overall LGTM

aggregator/cmd/main.go
emlautarom1: After #211 we no longer require to ask the aggregator for the `prometheus.Registry` since we can technically pass it as dependency: currently this is not done but it's a single change to a constructor parameter since the hard work has already been done.

Regarding the question of starting or not the servers in the aggregator: if the aggregator does not need a handle to the servers explicitly then it should **not** be responsible from instantiating them. I don't have enough context on the communication between the aggregator and the servers but if the only thing that they share is this registry then on main we should initialize it, and pass it to the three objects: `aggregator`, `restServer` and `rpcServer`. The less things the aggregator do the better

Hyodar: Yeah, as mentioned today, I don't have a problem with it right now, it's fine. What bothers me with initializing the servers outside is lifetime handling - we're passing an aggregator that could be closed

aggregator/aggregator.go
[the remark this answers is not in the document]
Hyodar: In `MessageBlsAggregationService` this means current block

aggregator/cmd/main.go
emlautarom1: > What mainly keeps bothering me with initializing the servers outside is lifetime handling - we're passing an aggregator that could be closed

That's a good point. `Aggregator` only gets closed through a context yet both servers are kept alive due to `http.ListenAndServe`. Ideally, all three services should be tied together with a single context but we cannot easily do that due to `http.ListenAndServe` not accepting a context (see: https://github.com/golang/go/issues/52805).

We need to implement our own graceful shutdown similar to what is described in this SO question: https://stackoverflow.com/questions/39320025/how-to-stop-http-listenandserve. This is outside the scope of this PR but it's worth revisiting.

aggregator/rest_server/server.go
```diff
@@ -0,0 +1,166 @@
...
+	s.logger.Info("Starting aggregator REST API.")
+
+	router := mux.NewRouter()
+	router.HandleFunc("/aggregation/state-root-update", wrapRequest(s.listener.APIErrors, s.handleGetStateRootUpdateAggregation)).Methods("GET")
+	router.HandleFunc("/aggregation/operator-set-update", wrapRequest(s.listener.APIErrors, s.handleGetOperatorSetUpdateAggregation)).Methods("GET")
+	router.HandleFunc("/checkpoint/messages", wrapRequest(s.listener.APIErrors, s.handleGetCheckpointMessages)).Methods("GET")
+
+	err := http.ListenAndServe(s.serverIpPortAddr, router)
+	if err != nil {
+		s.logger.Fatal("ListenAndServe", "err", err)
```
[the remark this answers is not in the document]
taco-paco: Fatal calls os.Exit

aggregator/rpc_server/server.go
```diff
@@ -0,0 +1,170 @@
...
+
+	err := rpc.RegisterName("Aggregator", s)
+	if err != nil {
+		s.logger.Fatal("Format of service TaskManager isn't correct. ", "err", err)
+	}
+	rpc.HandleHTTP()
+	err = http.ListenAndServe(s.serverIpPortAddr, nil)
+	if err != nil {
+		s.logger.Fatal("ListenAndServe", "err", err)
+	}
```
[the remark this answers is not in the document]
taco-paco: [Fatal calls os.Exit](https://github.com/NethermindEth/near-sffl/pull/204#discussion_r1631399860)

aggregator/cmd/main.go
taco-paco: Aggregator doesn't shutdown servers anyway rn. Shall be done on context.Doen similar to what @emlautarom1 linked [here](https://stackoverflow.com/questions/39320025/how-to-stop-http-listenandserve)

taco-paco: Resolving this. Will create another PR on `registry` and server shutdown matter

aggregator/cmd/main.go
taco-paco: Done

aggregator/rest_server/server.go
```diff
@@ -0,0 +1,166 @@
...
+
+	blockHeight, err := strconv.ParseUint(params.Get("blockHeight"), 10, 64)
+	if err != nil {
+		http.Error(w, "Invalid blockHeight", http.StatusBadRequest)
+		return err
+	}
+
+	response, err := s.app.GetStateRootUpdateAggregation(uint32(rollupId), blockHeight)
+	if err != nil {
+		http.Error(w, err.Error(), mapErrorToCode(err))
```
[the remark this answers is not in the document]
taco-paco: Done

emlautarom1 (approved):
LGTM

PR #208, opened 2024-06-04
Fix flaky `SafeEthClient` tests

### LINKED ISSUE #207: Flaky `SafeEthClient` tests
Our current unit tests are prone to fail on CI inconsistently (flaky tests). For example: https://github.com/NethermindEth/near-sffl/actions/runs/9303427652/job/25605715036#step:4:198

We would like to double check these tests to see what's wrong and fix them.

emlautarom1:
Fixes #207 by adding locks when needed and proper channel flushing. Issues were detected by using the `-race` flag. Only test code was changed.


Hyodar (approved):
LGTM! Thanks for that!

PR #209, opened 2024-06-05
Limit max DB connections to 1

Hyodar:
We were having an issue on DB management: whenever the operators were making a checkpoint, they were calling `GetAggregatedCheckpointMessages`, which triggered a `SELECT` on the DB. Since we are continuously writing to the DB as well, if a write happened to be happening at the same time, the DB would be locked, as it's SQLite - not supposed to be concurrent.
An immediate fix to this is simply limiting the connections, as effectively we are not using more than one at once.


taco-paco (approved):

Hyodar:
Merging this since it was already discussed with @emlautarom1 before.

PR #211, opened 2024-06-05
Flexibilize initialization on `BuildAll`

emlautarom1:
There are several usages of `eigenclients.BuildAll`:

https://github.com/NethermindEth/near-sffl/blob/43f66b473139d439f9695e2d73210da072e37447/plugin/cmd/operator_plugin.go#L90

https://github.com/NethermindEth/near-sffl/blob/43f66b473139d439f9695e2d73210da072e37447/operator/operator.go#L127

https://github.com/NethermindEth/near-sffl/blob/43f66b473139d439f9695e2d73210da072e37447/aggregator/aggregator.go#L122

https://github.com/NethermindEth/near-sffl/blob/43f66b473139d439f9695e2d73210da072e37447/tests/integration/registration_test.go#L87

This API is a bit cumbersome to use mainly due to not allowing to pass instances of `eth.Client` or disabling Prometheus metrics. It could be argued that we don't need this API at all since it wraps multiple constructors and instead we should build only what is immediately needed, avoiding in most cases duplicated work: 

https://github.com/Layr-Labs/eigensdk-go/blob/33241d37ea3a4d3ed3cfbb01010a3c70cad494d0/chainio/clients/builder.go#L33-L48

Nevertheless, this PR creates a similar API `chainio.BuildAll` that allows users to pass instances of `eth.Client` for HTTP and WS connections, and does not deal with metrics all together.

---

Surprisingly, there seems to be a "bug" in the version that we're using of the eigen-sdk where the `metrics.Metrics` are completely ignored during the construction of all dependencies:

- Creation:

https://github.com/Layr-Labs/eigensdk-go/blob/33241d37ea3a4d3ed3cfbb01010a3c70cad494d0/chainio/clients/builder.go#L58-L59

- Passed when constructing a `ElChainWriter`

https://github.com/Layr-Labs/eigensdk-go/blob/33241d37ea3a4d3ed3cfbb01010a3c70cad494d0/chainio/clients/builder.go#L179

- And ignored in the constructor

https://github.com/Layr-Labs/eigensdk-go/blob/33241d37ea3a4d3ed3cfbb01010a3c70cad494d0/chainio/clients/elcontracts/writer.go#L58


aggregator/aggregator.go
```diff
@@ -111,45 +111,41 @@ var _ core.Metricable = (*Aggregator)(nil)
...
-	if err != nil {
-		logger.Error("Cannot create sdk clients", "err", err)
-		return nil, err
+	withOptionalMetrics := func(url string) safeclient.SafeEthClientOption {
+		if !config.EnableMetrics {
+			return func(*safeclient.SafeEthClient) {}
+		}
+
+		registry := prometheus.NewRegistry()
+		rpcCallsCollector := rpccalls.NewCollector(AggregatorNamespace+url, registry)
```
emlautarom1, line 120: On each usage of `withOptionalMetrics` we're creating a new `registry` to maintain the previous behavior

core/chainio/builder.go
```diff
@@ -0,0 +1,214 @@
+package chainio
+
+// Rewrite of 'github.com/Layr-Labs/eigensdk-go/chainio/clients/builder.go' to be more flexible
```
emlautarom1, line 3: Specifically, this version: https://github.com/Layr-Labs/eigensdk-go/blob/33241d37ea3a4d3ed3cfbb01010a3c70cad494d0/chainio/clients/builder.go#L1

core/chainio/builder.go
```diff
@@ -0,0 +1,214 @@
...
+		ElChainReader:              elChainReader,
+		ElChainWriter:              elChainWriter,
+		AvsRegistryChainReader:     avsRegistryChainReader,
+		AvsRegistryChainSubscriber: avsRegistryChainSubscriber,
+		AvsRegistryChainWriter:     avsRegistryChainWriter,
+	}, nil
+
+}
+
+func buildElClients(
```
emlautarom1, line 96: It could be argued that we should expose these functions instead of `BuildAll` to allow for more granularity. For now we'll keep the currently used API

core/chainio/builder.go
```diff
@@ -0,0 +1,214 @@
...
+		ethHttpClient,
+		logger,
+		nil,
+		txMgr,
+	)
+
+	return elChainReader, elChainWriter, nil
+}
+
+func buildAvsClients(
```
emlautarom1, line 159: Same as in `buildElClients`

operator/avs_manager.go
```diff
@@ -92,8 +91,8 @@ func NewAvsManager(config *optypes.NodeConfig, ethRpcClient eth.Client, ethWsCli
 		avsReader:                    avsReader,
 		avsWriter:                    avsWriter,
 		avsSubscriber:                avsSubscriber,
-		eigenlayerReader:             sdkClients.ElChainReader,
-		eigenlayerWriter:             sdkClients.ElChainWriter,
+		eigenlayerReader:             elChainReader,
+		eigenlayerWriter:             elChainWriter,
```
emlautarom1, line 95: Instead of passing a large structure we pass only what we need.

operator/operator.go
```diff
@@ -116,36 +117,45 @@ func NewOperatorFromConfig(c optypes.NodeConfig) (*Operator, error) {
...
 	}
 
-	chainioConfig := clients.BuildAllConfig{
-		EthHttpUrl:                 c.EthRpcUrl,
-		EthWsUrl:                   c.EthWsUrl,
-		RegistryCoordinatorAddr:    c.AVSRegistryCoordinatorAddress,
-		OperatorStateRetrieverAddr: c.OperatorStateRetrieverAddress,
-		AvsName:                    AVS_NAME,
-		PromMetricsIpPortAddress:   c.EigenMetricsIpPortAddress,
+	reg := prometheus.NewRegistry()
```
emlautarom1, line 120: In this case we want to share the registry across both HTTP and WS clients to preserve the previous behavior.

operator/operator.go
```diff
@@ -116,36 +117,45 @@ func NewOperatorFromConfig(c optypes.NodeConfig) (*Operator, error) {
...
+		c.AVSRegistryCoordinatorAddress,
+		c.OperatorStateRetrieverAddress,
+		ethHttpClient,
+		ethWsClient,
+		ecdsaPrivateKey,
+		logger,
+	)
 	if err != nil {
-		return nil, err
+		panic(err)
```
emlautarom1, line 151: Do not get confused with the diff: previously we were already panicking of we could not build `sdkClients`

operator/operator.go
```diff
@@ -189,17 +204,28 @@ func NewOperatorFromConfig(c optypes.NodeConfig) (*Operator, error) {
...
-		logger:     logger,
-		ethClient:  ethRpcClient,
-		metricsReg: reg,
-		// TODO: replace with noop in case not enabled
-		metrics:                    sdkClients.Metrics,
+		config:                     c,
+		logger:                     logger,
+		ethClient:                  ethHttpClient,
+		metricsReg:                 reg,
+		metrics:                    optionalMetrics,
```
emlautarom1, line 228: Solves a pending TODO by only constructing relevant metrics when they're enabled in config.

tests/integration/registration_test.go
```diff
@@ -75,16 +75,24 @@ func TestRegistration(t *testing.T) {
...
 	}
 
-	buildClientConfig := sdkclients.BuildAllConfig{
-		EthHttpUrl:                 mainnetAnvil.HttpUrl,
-		EthWsUrl:                   mainnetAnvil.WsUrl,
-		RegistryCoordinatorAddr:    nodeConfig.AVSRegistryCoordinatorAddress,
-		OperatorStateRetrieverAddr: nodeConfig.OperatorStateRetrieverAddress,
-		AvsName:                    "super-fast-finality-layer",
-		PromMetricsIpPortAddress:   "127.0.0.1:9090",
+	ethHttpClient, err := eth.NewClient(mainnetAnvil.HttpUrl)
```
emlautarom1, line 78: The tests were not using `eth.SafeClient`. Do we want to test with it @Hyodar ?

aggregator/aggregator.go
```diff
@@ -111,45 +111,41 @@ var _ core.Metricable = (*Aggregator)(nil)
...
-	registry := clients.PrometheusRegistry
 
-	ethHttpClient, err := core.CreateEthClientWithCollector(
-		AggregatorNamespace,
-		config.EthHttpRpcUrl,
-		config.EnableMetrics,
-		registry,
-		logger,
-	)
+	ethHttpClient, err := safeclient.NewSafeEthClient(config.EthHttpRpcUrl, logger, withOptionalMetrics(config.EthHttpRpcUrl))
```
Hyodar, line 124: You could just use `core.CreateEthClientWithCollector` here IMO

Hyodar: Actually, you could also refactor that to treat a `nil` registry as a metrics disable

taco-paco:
That is a PR with good intentions. 

As mentioned yesterday in other thread [here](https://github.com/NethermindEth/near-sffl/pull/204#discussion_r1628009911) there was PR exposing and changing API of eigensdk.

I would advocate within this PR to raise the version of eigensdk to include those changes. and then change the logic accordingly.

They became available in [version v0.1.7](https://github.com/Layr-Labs/eigensdk-go/releases/tag/v0.1.7).  It worth taking a look if we can upgrade without breaking compatibility and with low effort

aggregator/aggregator.go
```diff
@@ -111,45 +111,41 @@ var _ core.Metricable = (*Aggregator)(nil)
...
-		registry,
-		logger,
-	)
+	ethWsClient, err := safeclient.NewSafeEthClient(config.EthWsRpcUrl, logger, withOptionalMetrics(config.EthWsRpcUrl))
 	if err != nil {
 		logger.Error("Cannot create ws ethclient", "err", err)
 		return nil, err
 	}
 
+	clients, err := chainio.BuildAll(
```
taco-paco, line 136: Within `chainio.BuildAll` you create `signerV2` & `txMgt` which are already created lower.

In this [PR](https://github.com/Layr-Labs/eigensdk-go/pull/117/files) I made `BuildAvsClient` & `BuildElClients` available for use. I would recommend using them instead of `BuildAll` and utilize `signerV2` & `txMgr` that are already created within this constructor

aggregator/aggregator.go
```diff
@@ -111,45 +111,41 @@ var _ core.Metricable = (*Aggregator)(nil)
...
-	clients, err := eigenclients.BuildAll(chainioConfig, config.EcdsaPrivateKey, logger)
-	if err != nil {
-		logger.Error("Cannot create sdk clients", "err", err)
-		return nil, err
+	withOptionalMetrics := func(url string) safeclient.SafeEthClientOption {
+		if !config.EnableMetrics {
+			return func(*safeclient.SafeEthClient) {}
+		}
+
+		registry := prometheus.NewRegistry()
```
taco-paco, line 119: Notice that you create registry here & on the line 230.

After in function `Aggregator::start`, specifically `agg.metrics.Start` call we use the latter one. Hence server doesn't scrape these ones.  

Aggregator shall accept registry as an argument, since creating registries is not responsibility of `Aggregator`.  It will also help with another PR where we splitted REST & RPC server from Aggregator.

taco-paco (changes requested):
Overall a great PR. But I would like to make some comments for imrpovment.

We shall stay close to eigensdk. What I mean. Even if updating rn will be a breaking change, I would propose to expose `buildAvsClients` & `buildElClients` as it done [here](https://github.com/Layr-Labs/eigensdk-go/blob/d9c2d1346a7c5b56afc7c6d2a39a2314a4ec1688/chainio/clients/builder.go#L221) and use them within constructor directly.
This is important once we actually decide to update version, so changing back to eigensdk would be pretty much seamless.

The second is registries & prometheus. I reckon that `registry` shall be passed to constructor of aggregator and not created within since aggregator isn't actually responsible for doing that.  Without this RPC & REST servers will have to ask aggregator for it as it currently done [here](https://github.com/NethermindEth/near-sffl/blob/0196995ae29790cc7a4f8a58f2bdb848ca58aae9/aggregator/cmd/main.go#L74) which is a hack and pretty ugly.

aggregator/aggregator.go
emlautarom1: Agree. This was a source of confusion when working in the PR. I'll refactor the code to not use `BuildAll` but rather what is exactly needed on each case avoiding duplicated work

aggregator/aggregator.go
emlautarom1: Good catch, fixed

aggregator/aggregator.go
emlautarom1: Integrated

emlautarom1:
@taco-paco

> We shall stay close to eigensdk. What I mean. Even if updating rn will be a breaking change, I would propose to expose buildAvsClients & buildElClients as it done [here](https://github.com/Layr-Labs/eigensdk-go/blob/d9c2d1346a7c5b56afc7c6d2a39a2314a4ec1688/chainio/clients/builder.go#L221) and use them within constructor directly.
This is important once we actually decide to update version, so changing back to eigensdk would be pretty much seamless.

The final solution ended up ignoring `buildAvsClients` and `buildElClients` using the appropriate constructors directly to avoid duplicated work. This allows each use case to only instantiate what is needed (ex. `AvsChainReader`, `ElChainWriter`, etc.) at the expense of having more duplicated initialization code. Nevertheless, this showed up that there were several call sites where we were constructing unnecessary objects (ex. `ecdsaPrivateKey` being constructed when it was actually not required)

> The second is registries & prometheus. I reckon that registry shall be passed to constructor of aggregator and not created within since aggregator isn't actually responsible for doing that. Without this RPC & REST servers will have to ask aggregator for it as it currently done [here](https://github.com/NethermindEth/near-sffl/blob/0196995ae29790cc7a4f8a58f2bdb848ca58aae9/aggregator/cmd/main.go#L74) which is a hack and pretty ugly.

I won't make this change in this PR since it's a bit out of scope (we wanted to remove the dependency on `BuildAll` in the first place), but the code now is in a good shape to introduce this refactor: change the `NewAggregator` to accept a registry as a parameter and use it instead of the local one.

operator/operator.go
```diff
@@ -159,25 +144,94 @@ func NewOperatorFromConfig(c optypes.NodeConfig) (*Operator, error) {
...
+		registryCoordinatorAddress,
+		logger,
+	)
 	if err != nil {
 		logger.Error("Cannot create AggregatorRpcClient. Is aggregator running?", "err", err)
 		return nil, err
 	}
 
-	avsManager, err := NewAvsManager(&c, ethRpcClient, ethWsClient, sdkClients, txMgr, logger)
+	avsManager, err := NewAvsManager(&c,ethHttpClient,ethWsClient,elChainReader,elChainWriter,txMgr,logger)
```
Hyodar, line 234: Is this properly formatted?

operator/operator.go
```diff
@@ -159,25 +144,94 @@ func NewOperatorFromConfig(c optypes.NodeConfig) (*Operator, error) {
...
+		elContractBindings.Slasher,
+		elContractBindings.DelegationManager,
+		elContractBindings.StrategyManager,
+		elContractBindings.StrategyManagerAddr,
+		elChainReader,
+		ethHttpClient,
+		logger,
+		nil,
+		txMgr,
+	)
```
Hyodar, line 221: Not sure how I feel about this - I get the intention of not constructing things that are not needed, but I feel like this may be getting a bit much. Maybe a method such as `createOperatorClients` or such would be interesting here.
Either way, not blocking.

Hyodar (approved):
LGTM! A few comments but not blocking.

operator/operator.go
emlautarom1: It wasn't. Fixed.

operator/operator.go
emlautarom1: I ended up encapsulating some of these constructors into more minimal APIs. The code now is much shorter and implementation details are hidden.

taco-paco (approved):
Looks pretty.

PR #243, opened 2024-06-06
Update eigensdk to 0.1.7

Hyodar:
This PR simply updates eigensdk to v0.1.7 and keeps compatibility.


aggregator/aggregator.go
```diff
@@ -325,7 +326,7 @@ func (agg *Aggregator) Close() error {
 func (agg *Aggregator) sendAggregatedResponseToContract(blsAggServiceResp blsagg.BlsAggregationServiceResponse) {
 	if blsAggServiceResp.Err != nil {
 		agg.aggregatorListener.IncErroredSubmissions()
-		if errors.Is(blsAggServiceResp.Err, blsagg.TaskExpiredError) {
+		if strings.Contains(blsAggServiceResp.Err.Error(), "expired") {
```
emlautarom1, line 329: Is there no support for `errors.Is` for this particular one?

emlautarom1 (approved):
LGTM

aggregator/aggregator.go
Hyodar: It's now an error function, so it's never the same error. This is honestly such a bummer, error handling should be easier in go. Not sure if there's a better way to do that, but seems unlikely

Hyodar:
Proceeding with this to make it easier for new PRs to use new eigensdk features / fixes.

taco-paco:
G move

Issue #240, opened 2024-06-06
Add healthcheck to operator container


Hyodar:

mikhailUshakoff:
I'm unsure whether we should include a health check inside the container. I don't see any benefits from it. Moving the logic from the health container to the operator container wouldn't change anything. Perhaps we should consider a more sophisticated health check algorithm.

Hyodar:
Hey, this is mostly a best practices thing. The container itself should have a general way for you to check if it`s healthy or not - you can always override it, but overall a separate service is not ideal. We did have a separate service before since our container had only the bare minimum infra.

PR #244, opened 2024-06-06
Add docker compose test

mikhailUshakoff:


Hyodar (approved):
LGTM!

.github/workflows/compose-test.yml
```diff
@@ -0,0 +1,126 @@
...
+              echo "$container is not running"
+              exit 1
+            fi
+          done
+
+
+      - name: Check health status
+        run: |
+          services=(
+            "near-sffl_operator1-health_1"
```
taco-paco, line 107: Is the intention here to check health of `operator` or `operator1-health` image? Because the latter is done. As I remember operator1-health will be running even if operator crashes

Hyodar: Discussed in call, `operator` should be checked for running state, and `operator-health` for healthiness until #240 is solved

PR #246, opened 2024-06-06
Add nil check for aggregation

### LINKED ISSUE #241: Add non-nil check to aggregation when fetching DB message
Currently, in all `database.go` methods, we're not checking for nil `Aggregation`s before doing `ToMessage` or such. We should do that.

mikhailUshakoff:
Fix  #241

### DIFF
--- a/aggregator/database/database.go
+++ b/aggregator/database/database.go
@@ -1,6 +1,7 @@
 package database
 
 import (
+	"context"
 	"errors"
 	"log"
 	"math"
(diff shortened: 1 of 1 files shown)


aggregator/database/database.go
```diff
@@ -288,13 +288,19 @@ func (d *Database) FetchCheckpointMessages(fromTimestamp uint64, toTimestamp uin
 	operatorSetUpdateMessageAggregations := make([]messages.MessageBlsAggregation, 0, len(operatorSetUpdates))
 
 	for _, stateRootUpdate := range stateRootUpdates {
+		if stateRootUpdate.Aggregation == nil {
```
Hyodar, line 291: I don't think we want to necessarily fail here. IMO we should just skip non-aggregated messages and log a warning for now.

taco-paco (approved):

Hyodar (approved):

emlautarom1 (approved):

Issue #239, opened 2024-06-06
Improve parameter validation in aggregator routes


Hyodar:
Basically in all API routes (REST and RPC) we are not validating parameters enough. For example, doing a non-nil check on the signed message signature or adding bounds to the checkpoint messages fetching.

emlautarom1:
> doing a non-nil check on the signed message signature

Could you elaborate on this?

emlautarom1:
> adding bounds to the checkpoint messages fetching

Added on #250

Hyodar:
> Could you elaborate on this?
The signed messages `.Signature` is a pointer, even though the signed message struct itself is a value.

PR #250, opened 2024-06-11
Server parameter validation

emlautarom1:
## Current Behavior

Partially solves https://github.com/NethermindEth/near-sffl/issues/239 by introducing additional bounds checks to the checkpoint messages fetching. It also adds additional tests for parameter deserializing.

## New Behavior

Ensures that ranges `from - to` are valid, that is `from < to` is always satisfied.

## Breaking Changes

Certain queries that would previously return an empty result set now return an error message indicating that the range provided is invalid.

## Observations

There is quite a complicated issue with the storage of `BlockNumber` and `Timestamp` as defined here:

https://github.com/NethermindEth/near-sffl/blob/18544a6716cb6fe0cb91f928ba1f476cc7bcaa5f/core/types/fields.go#L14-L15

> I'm assuming that this timestamp is representing a Unix timestamp

These types are used in `StateRootUpdateMessage`

https://github.com/NethermindEth/near-sffl/blob/18544a6716cb6fe0cb91f928ba1f476cc7bcaa5f/core/types/messages/state_root_update.go#L18-L19

This message type then gets converted to a specific model when interacting with the DB:

https://github.com/NethermindEth/near-sffl/blob/18544a6716cb6fe0cb91f928ba1f476cc7bcaa5f/aggregator/database/models/state_root_update.go#L12-L13

As noted, there is a pending range check to perform. SQLite by default does not support `uint64` values (and neither other DBs like Postgres), allowing only `int64` values. For both fields, if we try to store `math.MaxUint64` we'll get the following SQL error:

```
uint64 values with high bit set are not supported
```

I'm not 100% sure who is to blame here (gorm, SQLite), but this means that in the future, if we have either a `BlockHeight` or a `Timestamp` greater than `math.MaxInt64` then we'll get an error.

Possible actions:
1. Do nothing since we can guess that such high values will not show up in the near future.
2. Add a check to prevent inserting these values in the DB. This is not that different from 1 but at least we control what to do on errors
3. Change the DB schema to store these values, ex. store them as `text`. Curiously we get the same error on `BlockHeight` despite technically using `text` as a type which makes things a bit confusing. Either way, changing the schema would mean a breaking change which we want to avoid at least for the time being.

### DIFF
--- a/aggregator/database/database.go
+++ b/aggregator/database/database.go
@@ -263,6 +263,10 @@ func (d *Database) FetchCheckpointMessages(fromTimestamp uint64, toTimestamp uin
 		return nil, errors.New("timestamp does not fit in int64")
 	}
 
+	if (toTimestamp < fromTimestamp) {
+		return nil, errors.New("toTimestamp is less than fromTimestamp")
+	}
+
 	start := time.Now()
 	defer func() { d.listener.OnFetch(time.Since(start)) }()
 
--- a/aggregator/database/database_test.go
+++ b/aggregator/database/database_test.go
@@ -313,3 +313,21 @@ func TestFetchCheckpointMessages(t *testing.T) {
 		OperatorSetUpdateMessageAggregations: []messages.MessageBlsAggregation{},
 	})
 }
+
+func TestFetchCheckpointMessages_TimestampTooLarge(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
+
+	db, err := database.NewDatabase(":memory:")
+	assert.Nil(t, err)
+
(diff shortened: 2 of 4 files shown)


mikhailUshakoff (approved):
There is no problem. 
A Timestamp represents the number of non-leap nanoseconds since January 1, 1970, 0:00:00 UTC. Using i64 , dates can be represented as nanoseconds between 1677-09-21T00:12:43.145224192 and 2262-04-11T23:47:16.854775807.
The BlockHeight increases at a rate no faster than one per nanosecond. Therefore, it will take a substantial amount of time before it surpasses the i64 limit.

emlautarom1:
We'll go with option 1. for now (`Do nothing since we can guess that such high values will not show up in the near future`)

PR #260, opened 2024-06-21
Update `make` based instructions

emlautarom1:
## Current Behavior

`make start-operator` errors out immediately with the following error:

```
{"err":"open tests/keys/bls/1.bls.key.json: no such file or directory"}
````

More details on #257

## New Behavior

This PR updates `make` based instructions to match the current `Docker` setup.

## Breaking Changes

None beside getting something to work that was previously broken.

## Observations

- Added a new `near-da-rpc-sys` make command that builds the native libraries required by the relayer. 
- Adjusted the `CGO` linker flags so the relayer can pick the required native libraries during build time.
- Updated `README` with instructions on how to build the native libraries.
- Updated `config-files/operator.anvil.yaml` to match the latest file structure regarding keys.

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -121,6 +122,13 @@ tests-unit: ## runs all unit tests
 tests-contract: ## runs all forge tests
 	cd contracts/evm && forge test
 
+near-da-rpc-sys:
+	rm -rf relayer/libs && \
+	mkdir relayer/libs && \
+	cd submodules/rollup-data-availability && \
```
mikhailUshakoff, line 128: I'm unable to run `make near-da-rpc-sys` because the `submodules` folder is missing on my machine. Should we add it as a Git submodule?

mikhailUshakoff (changes requested):

Makefile
emlautarom1: Indeed, it seems like I got confused and thought that repo was already a submodule.

Makefile
```diff
@@ -104,7 +104,8 @@ start-indexer: ##
 	cargo run -p indexer --release -- --home-dir ~/.near/localnet run --da-contract-ids da.test.near --rollup-ids 2 --rmq-address "amqp://127.0.0.1:5672"
 
 start-test-relayer: ##
-	go run relayer/cmd/main.go --rpc-url ws://127.0.0.1:8546 --da-account-id da.test.near
+	export CGO_LDFLAGS="-L ./relayer/libs"
```
taco-paco, line 107: It's more complicated than that [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/gopkg/da-rpc/near_darwin.go#L5). The setup is platform dependent

And it also seems that it will override those flags

taco-paco (changes requested):
I would propose to refrain from getting rollup-data-availability as submodule. 

We shall list rollup-data-availability artifacts as prerequisites. Maybe with explanation on how to compile it.

The compilation of relayer is platform-dependant and rollup-data-availability handles that [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/gopkg/da-rpc/near_darwin.go#L5) and [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/gopkg/da-rpc/near_unix.go#L5).


User shall invoke make command [here](compile it using this [make](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/Makefile#L37) and then move it to /usr/local/lib. Go will pick it up itself after that

Makefile
taco-paco: This shall be defined on a particular package level

taco-paco: From doc [here](https://pkg.go.dev/cmd/cgo)

> When building, the CGO_CFLAGS, CGO_CPPFLAGS, CGO_CXXFLAGS, CGO_FFLAGS and CGO_LDFLAGS environment variables are added to the flags derived from these directives. Package-specific flags should be set using the directives, not the environment variables, so that builds work in unmodified environments. Flags obtained from environment variables are not subject to the security limitations described above.

 this my statement is incorrect

> Also that seems dangerous since other libraries may also have their own CGO_LDFLAGS definitions and may fail to compile

Nevertheless documentation still points out the same thing:

> Package-specific flags should be set using the directives, not the environment variables, so that builds work in unmodified environments.

firatsertgoz: May I suggest another angle,

How about instead of using `da-rpc-sys` you straight up use the `http-api`. Yes you would need to run a sidecar with the relayer, but the seperation would be very clean.

firatsertgoz:
May I suggest another angle,

How about instead of using `da-rpc-sys` you straight up use the `http-api`. Yes you would need to run a sidecar with the relayer, but the seperation would be very clean.

taco-paco:
> May I suggest another angle,
> 
> How about instead of using `da-rpc-sys` you straight up use the `http-api`. Yes you would need to run a sidecar with the relayer, but the seperation would be very clean.

That is a great idea. @emlautarom1 do you want to play with it? This would require small Rust service on top of rollup-data-availability.

Makefile
emlautarom1: > Also that seems dangerous since other libraries may also have their own CGO_LDFLAGS definitions and may fail to compile

To avoid overriding the environment we could do something like `CGO_LDFLAGS="-L ./relayer/libs go run ...", but the fact that the flags are platform dependent makes things a bit tricky.

Nevertheless, lets follow the docs: we'll set package specific flags through directives.

emlautarom1:
> I would propose to refrain from getting rollup-data-availability as submodule.
> 
> We shall list rollup-data-availability artifacts as prerequisites. Maybe with explanation on how to compile it.
> User shall invoke make command [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/Makefile#L37) and then move it to /usr/local/lib. Go will pick it up itself after that

Initially I worked on a different approach that relied on cloning the `rollup-data-availability` repo, run the appropriate make command, copied the artifacts and then deleted the repo. This removes the need for the submodule and ensures that we're building for the correct platform, without introducing additional friction.

I'm against installing the libraries globally (ex. `/usr/local/lib`), so we would like to "install" them locally (this PR uses `relayer/libs`). The issue is then how do we ensure that the Go compiler includes that path during the linking phase without polluting the global defaults.

I think defining the make command to something like `CGO_LDFLAGS="-L ./relayer/libs" go run ...` should be enough.

> The compilation of relayer is platform-dependant and rollup-data-availability handles that [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/gopkg/da-rpc/near_darwin.go#L5) and [here](https://github.com/taco-paco/rollup-data-availability/blob/e6354bde000692b84a4dde1581babe2619b5b00e/gopkg/da-rpc/near_unix.go#L5).

My understanding is that our relayer (`relayer/`) is not platform dependent but rather it has a dependency on native libraries. What we need to do is to point Go to the correct directory while compiling the relayer to properly pick up these libraries.

emlautarom1:
Regarding the alternative proposed by @firatsertgoz, I think it would be nice to explore it later. For now I would like to fix the current `make` based instructions without adding additional dependencies, in particular considering that the `docker-compose` workflow is already working as intended (and it uses `da-rpc-sys` under the hood).

Hyodar:
About the Go sidecar conversation - I agree with both. We should move to the Go sidecar but IMO this PR should keep its scope.

Reason for that is already in #165. We can move ahead with it, but I'd still prefer if it kept a compatibility layer in this case. AFAIK we don't have other breaking updates piled up and coordinating those should be very costly - there are nodes from all over the world and some are more responsive than others.

taco-paco:
@Hyodar thanks for implicitly reminding about the PR)
With regards to sidecar we can just use the current commit of near-da, thus it shall be only relayer update. After we would probably be able to discard #165 and just update the sidecar along with operator.

@emlautarom1 
> I'm against installing the libraries globally (ex. /usr/local/lib), so we would like to "install" them locally (this PR uses relayer/libs). The issue is then how do we ensure that the Go compiler includes that path during the linking phase without polluting the global defaults.

The thing is that `CGO_LDFLAGS` shall be customizable by user himself not by us. There're multiple C libraries that our code depends on implicitly handled by go itself we can't specify them all for users. If user has custom locations with zlib, glibc, rollup-da he can use `CGO_LDFLAGS` himself.

Hyodar:
Okay, all conversation aside, I think the first question we should answer here is whether this tooling is important atm. It was used before and increasingly unused over time. Are we using these individual commands, or should we just deprecate this and use exclusively docker compose containers? On my end, I usually just use the containers directly, as it's actually closer to the real scenario.

emlautarom1:
> Are we using these individual commands, or should we just deprecate this and use exclusively docker compose containers? On my end, I usually just use the containers directly, as it's actually closer to the real scenario.

I favor deprecating the `make` based instructions and only use `docker` as the unique workflow (less maintenance burden) but that's outside the scope of this PR. Note that we should still have `make` for development purposes (ex. run tests, build images, etc.)

Hyodar:
Okay, after checking the whole conversation again, I do think we can generally move forward with this. My overall take is:
* Depending on a global library to be managed by the user is not nice as this is a quite specific application. We should try to keep things local as much as possible.
* This is going to be deprecated (really) soon, I don't think we should spend more headspace with this

That said, I understand the trouble here, but overall I think defining the library as it is the best approach for now, though I'd highly prefer if we concatted (e.g. `CGO_LDFLAGS += -L/path/to/your/library`) instead of redefining the env variable. This way the user is not tied to our definition. The compilation instructions should also be updated in the docs.

Again, this should be deprecated really soon in favor of only container-based testing, so we don't have much to gain on this discussion.

emlautarom1:
> That said, I understand the trouble here, but overall I think defining the library as it is the best approach for now, though I'd highly prefer if we concatted (e.g. CGO_LDFLAGS += -L/path/to/your/library) instead of redefining the env variable. This way the user is not tied to our definition

Done, now we concatenate existing `CGO_LDFLAGS` if any, allowing the user to configure this behavior if needed.

> The compilation instructions should also be updated in the docs.

Already updated the main `README.md` with new instructions as described in the PR body.

Hyodar (approved):
LGTM! Please update the `docs/` intro page as well.

Makefile
emlautarom1: Solved by reusing existing `CGO_LDFLAGS`. Closing discussion to merge.

Hyodar:
Added the deprecation of the non-container testing environment instructions as #269.

mikhailUshakoff (approved):

emlautarom1:
Merging as discussed during today's daily

Issue #214, opened 2024-06-05
Integrate custom BLS aggregation service to checkpoint tasks


Hyodar:
Currently, we are using `eigen-sdk`'s BLS aggregation service for task response aggregations. We should move this to our custom implementation (currently it's named specifically for messages, but we should have something similar for tasks) - the benefits are considerable, being the most notable of them the ability to differentiate reaching quorum from a full quorum. After this is done, the checkpoint aggregation quorum threshold should be changed to `66`.

Hyodar:
Solved in #282

PR #263, opened 2024-06-28
Custom BLS aggregation service for Checkpoint Tasks

emlautarom1:
## Current Behavior

See #214 for details

## New Behavior

We introduce a `TaskBlsAggregatorService` that replaces EigenLayer's SDK [BlsAggregationService](https://github.com/Layr-Labs/eigensdk-go/blob/cc3b8d2dd3390ce407aadc76130c4b3fa583f9b5/services/bls_aggregation/blsagg.go) specifically for checkpoint tasks.

## Breaking Changes

There should be no breaking changes

## Observations

There is a lot of what seems to be duplicated code between `TaskBlsAggregatorService` and `MessageBlsAggregatorService`. I've spent quite some time trying to reconcile these differences but I just cannot understand the modules well enough.

The main differences are that `Messages` work with `MessageDigest`, while `Tasks` use `TaskIndex`. Also, `TaskBlsAggregation[ServiceResponse]` and `MessageBlsAggregation[ServiceResponse]` look very similar, but not quite enough.


Hyodar:
Looks good. Then, as mentioned in #214, we should integrate the same partial aggregation functionality present in MessageBlsAggregationService to this as well. IMO the rationale here is checking what specific parts should be changed and how much complexity it adds.
It may just be better (esp as Go doesn't have many typing features) just making two separate modules with some duplicated code than trying to come up with a generalized interface and making everything more painful to deal with if specific changes are required.

emlautarom1:
> Looks good. Then, as mentioned in https://github.com/NethermindEth/near-sffl/issues/214, we should integrate the same partial aggregation functionality present in MessageBlsAggregationService to this as well.

I would like to do that on a separate PR to keep the scope of this one small.

> It may just be better (esp as Go doesn't have many typing features) just making two separate modules with some duplicated code than trying to come up with a generalized interface and making everything more painful to deal with if specific changes are required.

I actually spent quite some time trying to get the `MessageBlsAggregationService` to work with tasks but the signatures of the interfaces and the result values are just different though maybe there is a way to unify them based on some domain knowledge that I lack. For the time being I agree on keeping the current duplicated code rather than trying to force a common interface where there is none.

Hyodar:
Found a good way of using the same module for tasks as well, so decided to work on this in #282 instead of here. Closing this now.

Issue #247, opened 2024-06-06
Include signature in message receiving logs


Hyodar:
We should also log the signatures from received signed task responses and messages - this way, we can also access the individual signature through logs for any debugging process. The current log displays the signature pointer only.
This was already implemented in a naive way in `testnet-0`, but should be revisited.

emlautarom1:
This is already implemented:

https://github.com/NethermindEth/near-sffl/blob/8f43707c78ba61820b3d0b18a27c2afe2e930672/aggregator/rpc_server/server.go#L95

https://github.com/NethermindEth/near-sffl/blob/8f43707c78ba61820b3d0b18a27c2afe2e930672/aggregator/rpc_server/server.go#L115

https://github.com/NethermindEth/near-sffl/blob/8f43707c78ba61820b3d0b18a27c2afe2e930672/aggregator/rpc_server/server.go#L133

For example, this is how we're logging `messages.SignedCheckpointTaskResponse`:

```
[...]

PR #264, opened 2024-07-01
Include signature in message receiving logs

emlautarom1:
## Current Behavior

Currently, when logging messages that contain a `bls.Signature` we're logging a pointer address rather than its contents.

## New Behavior

Whenever we log these messages we log their value rather that their representation

## Breaking Changes

This change should not introduce any breaking changes.

## Observations

The original code used `fmt.Sprintf("%#v", ...)` in the log calls which did not dereference the pointers. This behavior is present since #12 but there is no clear indication on why it was done that way.

### DIFF
--- a/aggregator/message_blsagg.go
+++ b/aggregator/message_blsagg.go
@@ -472,7 +472,7 @@ func (mbas *MessageBlsAggregatorService) verifySignature(
 ) error {
 	_, ok := operatorsAvsStateDict[signedMessageDigest.OperatorId]
 	if !ok {
-		mbas.logger.Warn("Operator not found. Skipping message", "operator", fmt.Sprintf("%#v", signedMessageDigest.OperatorId))
+		mbas.logger.Warn("Operator not found. Skipping message", "operator", signedMessageDigest.OperatorId)
 		return OperatorNotPartOfMessageQuorumErrorFn(signedMessageDigest.OperatorId, signedMessageDigest.MessageDigest)
 	}
 
--- a/aggregator/rpc_server/server.go
+++ b/aggregator/rpc_server/server.go
@@ -2,7 +2,6 @@ package rpc_server
 
 import (
 	"errors"
-	"fmt"
 	"net/http"
 	"net/rpc"
 	"strings"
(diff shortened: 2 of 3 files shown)


Hyodar:
Great! Can you include a snippet of the logs right now?

mikhailUshakoff (approved):

emlautarom1:
It will look something like this:

```
2024-07-01T13:31:40.252-0300    INFO    rest_server/log_test.go:23      Received signed operator set update message     {"message": {"Message":{"Id":100,"Timestamp":200,"Operators":[{"Pubkey":{"X":3,"Y":4},"Weight":5}]},"BlsSignature":{"g1_point":{"X":1,"Y":2}},"OperatorId":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}}
```

I don't see any `ToHexString` or similar in the EigenLayer SDK for `bls.Signature`. The closest thing is `G1Point.String()` which shows `"E([1,2])"` (not much better in my opinion)

emlautarom1:
It will look something like this:

```
2024-07-01T13:31:40.252-0300    INFO    rest_server/log_test.go:23      Received signed operator set update message     {"message": {"Message":{"Id":100,"Timestamp":200,"Operators":[{"Pubkey":{"X":3,"Y":4},"Weight":5}]},"BlsSignature":{"g1_point":{"X":1,"Y":2}},"OperatorId":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}}
```

To get a better representation we could use something like `BlsSignature.Serialize()` which looks like the following:

```
2024-07-01T13:33:32.477-0300    INFO    rest_server/log_test.go:23      Received signed operator set update message     {"message": {"Message":{"Id":100,"Timestamp":200,"Operators":[{"Pubkey":{"X":3,"Y":4},"Weight":5}]},"BlsSignature":{"g1_point":{"X":1,"Y":2}},"OperatorId":[0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0]}, "signature": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAEAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAg=="}
```

Hyodar (approved):

Hyodar:
That's all right really. Usually these signatures are represented as `E(x, y)` indeed, without hex.

emlautarom1:
@Hyodar do you want to include the `Serialize` version or is it enough with the `X/Y` (default) format?

emlautarom1:
Merging and closing #247

PR #265, opened 2024-07-01
Nffl demo examples

mikhailUshakoff:
## Current Behavior

Currently, there are no examples demonstrating how to interact with our protocol.

## New Behavior

Implement simple scripts to:
- Update the operator set
- Update the state root
- Get storage value

## Breaking Changes

None


Hyodar:
GJ! I know this is not yet open for reviews, just wanted to add something already.
Wdyt about making this into simple CLI programs? Should not be too hard, and we can specify some parameters as flags. I'm thinking like `./updateStateRoot.js --from arbitrum --to optimism --private-key <>`. We can have a centralized config that reads the addresses from the scripts output folder. I think this would be the logical step here.

Hyodar:
I'm a bit familiar with JS CLI tooling, so maybe it's more efficient if I get to it instead

demo-nffl-cli/updateOperatorSet.js
```diff
@@ -0,0 +1,62 @@
...
+    if (!response.ok) {
+        throw new Error(`HTTP error! status: ${response.status}`);
+    }
+    const respText = await response.text();
+    text = respText.replace(/"Weight":\s*(\d+)/g, '"Weight": "$1"');
+    data = JSON.parse(text);
+    const operators = data.Message.Operators.map(({ Pubkey, Weight }) => ({
+        pubkey: Pubkey,
+        weight: Weight
+    }));
```
Hyodar, line 31:
```suggestion
    const operators = data.Message.Operators.map(({ Pubkey: pubkey, Weight: weight }) => ({ pubkey, weight });
```

demo-nffl-cli/updateOperatorSet.js
```diff
@@ -0,0 +1,62 @@
...
+    // Create signature info
+    const signatureInfo =  {
+        nonSignerPubkeys,
+        apkG2: {
+            X: [data.Aggregation.SignersApkG2.X.A1,data.Aggregation.SignersApkG2.X.A0],
+            Y: [data.Aggregation.SignersApkG2.Y.A1,data.Aggregation.SignersApkG2.Y.A0]
+        },
+        sigma: data.Aggregation.SignersAggSigG1.g1_point
+    }
+    // Call contact
```
Hyodar, line 54:
```suggestion
    // Call contract
```

demo-nffl-cli/getStorageValue.js
```diff
@@ -0,0 +1,66 @@
...
+ * from the source chain and verifies it on the destination chain
+*/
+async function getStorageValue(options) {
+    //
+    // Get proof
+    //
+    const srcProvider = new ethers.JsonRpcProvider(options.srcRpcUrl);
+    //chainId
+    const { chainId } = await srcProvider.getNetwork();
+    // Prepear params
```
Hyodar, line 17:
```suggestion
    // Prepare params
```

Hyodar (commented):
Only small nit-picks, overall LGTM! I'll just ask you to check these comments and set up a small formatter (should be pretty easy) with an `npm run format` command - there are one or two formatting issues, so it's easier to just set one up.

firatsertgoz:
fyi - I am turning this into a very lean typescript library so people can import it in their projects. WIll link the repo later.

firatsertgoz:
Shall we merge this? It would help applications who are integrating to have a base example. @Hyodar what do you think?

Hyodar:
Hey! We could, let me take a look at this again. Still, if it's being moved to a library, IMO it's better to have this using the library already.

firatsertgoz:
Lets merge the example first, I'll create a PR with the library later on.

Hyodar:
Alright, I'll do some changes there and will merge it soon

PR #255, opened 2024-06-17
Replace all occurrences of SFFL with NFFL

emlautarom1:
## Current Behavior

Solves #254 by applying the proper renaming in all docs

## New Behavior

SFFL is now called NFFL (`Nuffle Fast Finality Layer`)

## Breaking Changes

There should be no breaking changes since all internal names (file names, enviroment variables, contracts, etc) still use the old `sffl` name

PR #266, opened 2024-07-01
Fix SFFL title in docs

emlautarom1:
## Current Behavior

The current docs (https://nffl.nethermind.io) still show `NEAR SFFL` in the page title and navbar.

## New Behavior

The title and navbar show `NFFL` rather than `SFFL` (fixes https://github.com/NethermindEth/near-sffl/pull/255#issuecomment-2200686227)

## Breaking Changes

There should be no breaking changes.

## Observations

We missed this on #255

### DIFF
--- a/docs/docusaurus.config.js
+++ b/docs/docusaurus.config.js
@@ -8,8 +8,8 @@ import {themes as prismThemes} from 'prism-react-renderer';
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
-  title: 'NEAR NFFL',
-  tagline: 'NEAR Fast Finality Layer',
+  title: 'NFFL',
+  tagline: 'Nuffle Fast Finality Layer',
   favicon: 'img/favicon.ico',
 
   url: 'https://nffl.nethermind.io',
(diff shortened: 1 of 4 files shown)


Hyodar:
LGTM, though IMO we should also replace `NEAR SFFL` -> `NFFL` and `NEAR Fast Finality Layer` -> `Nuffle Fast Finality Layer`

docs/docusaurus.config.js
```diff
@@ -8,8 +8,8 @@ import {themes as prismThemes} from 'prism-react-renderer';
 
 /** @type {import('@docusaurus/types').Config} */
 const config = {
-  title: 'NEAR SFFL',
-  tagline: 'NEAR Super Fast Finality Layer',
+  title: 'NEAR NFFL',
+  tagline: 'NEAR Fast Finality Layer',
```
mikhailUshakoff, line 12:
```suggestion
  tagline: 'Nuffle Fast Finality Layer',
```

mikhailUshakoff (dismissed):

docs/docusaurus.config.js
emlautarom1: Good catch, fixed.

Hyodar (approved):

PR #267, opened 2024-07-02
Fix docker compose file

mikhailUshakoff:
## Current Behavior

The `docker compose up` command currently fails during the startup of the `indexer-setup` container. As a result, the `relayer` container do not run, leading to incorrect system operation. This failure occurs because the `indexer-setup` container attempts to deploy contracts that already exist in initial state.

## New Behavior

Remove the `indexer-setup` container entirely from the system.

## Breaking Changes

No breaking changes are expected.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,8 +81,6 @@ services:
     depends_on:
       rollup0-anvil:
         condition: service_healthy
-      indexer-setup:
-        condition: service_completed_successfully
     expose:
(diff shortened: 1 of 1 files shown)


Hyodar:
As discussed, the issue in this case is the volumes aren't cleared on each run, so the NEAR CLI data already exists even though the localnet is reset. Closing this for now and I should open a new PR soon.

PR #272, opened 2024-07-03
Clear NEAR DA data on docker compose test env reset

Hyodar:
## Current Behavior

NEAR DA data was not being cleared on a docker compose test env reset. The usual fix for this would be resetting the volumes whenever resetting - `docker compose down -v` or such. Still, this may pass unnoticed and cause some confusion.

A related PR is #267.

## New Behavior

The above issue is prevented by clearing the relevant data on the `indexer_setup` service. So, when re-upping it should be clean even if the volumes were not reset.

## Breaking Changes

None

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,7 +81,7 @@ services:
     depends_on:
       rollup0-anvil:
         condition: service_healthy
-      indexer-setup:
+      near-da-deploy:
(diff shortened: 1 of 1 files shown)


mikhailUshakoff (approved):
Well done!

emlautarom1 (approved):
:+1:

PR #273, opened 2024-07-03
Aggregator message TTL

emlautarom1:
## Current Behavior

There is no submission timeout for messages being sent to the aggregator, meaning that we spend time processing messages that are too old. See #212 for more details.

## New Behavior

We introduce a timeout based on the Aggregator current time (Unix timestamp), rejecting messages that are too old. We also adjust the checkpoint time range logic to be `(lastCheckpointTimestamp + 1) to (currentTime - TIMEOUT)`

## Breaking Changes

There is a risk of now dropping messages if the timeouts are too strict. We'll need to monitor this change once its deployed.

## Observations

The initial implementation uses a 60 second timeout. We also introduce a `Clock` abstraction for ease of testing that could be used in other parts of the codebase. It's up to discussion if we want to introduce a third-party library with more features to cover this use case.


aggregator/types/types.go
```diff
@@ -18,6 +18,7 @@ const QUERY_FILTER_FROM_BLOCK = uint64(1)
 
 const MESSAGE_TTL = 1 * time.Minute
 const MESSAGE_BLS_AGGREGATION_TIMEOUT = 30 * time.Second
+const MESSAGE_SUBMISSION_TIMEOUT = 1 * time.Minute
```
emlautarom1, line 21: What would be a reasonable submission timeout?

aggregator/aggregator.go
```diff
@@ -393,7 +395,7 @@ func (agg *Aggregator) sendNewCheckpointTask() {
 		fromTimestamp = 0
 	}
 
-	toTimestamp := block.Time()
+	toTimestamp := block.Time() - uint64(types.MESSAGE_SUBMISSION_TIMEOUT.Seconds())
```
Hyodar, line 398: This should also include the aggregation timeout (see #212). It makes sense because worst-case is aggregation + submission timeout.

aggregator/types/types.go
Hyodar: This should be okay

core/utils.go
```diff
@@ -86,3 +87,9 @@ func ConvertBytesToQuorumNumbers(input []byte) []eigentypes.QuorumNum {
 	}
 	return output
 }
+
+type Clock struct {
+	Now func() time.Time
+}
+
+var SystemClock = &Clock{Now: time.Now}
```
Hyodar, line 95: Why have this as a pointer?

Hyodar (approved):
LGTM! A few comments only.

core/utils.go
emlautarom1: I think it's a leftover of my first approach which used interfaces. Changing it to a value.

aggregator/aggregator.go
emlautarom1: Adjusted to include `- uint64(types.MESSAGE_BLS_AGGREGATION_TIMEOUT.Seconds())`

core/utils.go
Hyodar: Didn't mind it too much at this point, but any specific reason you changed the approach?

emlautarom1: Less code honestly. If we were to use an interface we would have:

```go
type Clock interface {
    Now() time.Time
}

type SystemClock struct { }
func (sc *SystemClock) Now() time.Time { return time.Now() }
var _ Clock = &SystemClock{}

type FixedClock struct {
    NowValue time.Time
}

func (fc *FixedClock) Now() time.Time { return fc.NowValue }
var _ Clock = &FixedClock{}
```

You really don't win anything over a `struct` that holds a function (final approach). If we wanted to use something more configurable I suggest looking for a proper "clock" library (ex. https://github.com/kubernetes/utils/blob/master/clock/clock.go)

Hyodar: Yup, that's okay for now

Hyodar (approved):

